PR #69, opened 2015-09-16
Move to xunit 2.1 RC1

ericeil:
The biggest pain here was that xunit is using a different form of Assembly.Load in RC1 (on accident, it seems - I've opened an issue in xunit).  This resulted in a failure to find the test assembly, when discovering tests in xunit.performance.run.  I've reworked things to work around the problem; this has a nice side-benefit of reducing the number of dependencies in most of our projects.

@ianhays, @pharring


src/xunit.performance.run/ProgramCore.cs
```diff
@@ -235,11 +239,10 @@ private IEnumerable<PerformanceTestInfo> DiscoverTests(IEnumerable<XunitProjectA
 
                 // Note: We do not use shadowCopy because that creates a new AppDomain which can cause
                 // assembly load failures with delay-signed or "fake signed" assemblies.
-                using (var controller = new Xunit2(
-                    useAppDomain: false,
-                    sourceInformationProvider: new NullSourceInformationProvider(),
+                using (var controller = new XunitFrontController(
                     assemblyFileName: assembly.AssemblyFilename,
                     shadowCopy: false,
```
ianhays: should shadowCopy be ShadowCopy/ShadowCopyOrDefault? I've only glanced at the changes xunit made, so I'm not sure

samples/ClassLibrary.net46/ClassLibrary.net46.csproj
```diff
@@ -65,6 +64,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="app.config" />
```
ianhays: wasnt this removed?

ianhays:
A few questions, but otherwise no comments. Thanks for doing this; LGTM.

src/xunit.performance.run/ProgramCore.cs
ericeil: I think this is still correct.

samples/ClassLibrary.net46/ClassLibrary.net46.csproj
ericeil: Thanks; fixed.

PR #71, opened 2015-09-22
More DNX work

ericeil:


ericeil:
@dotnet-bot test this please

dnx/xunit.performance.run.dnx/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "profiles": {
+    "xunit.performance.run.dnx": {
+      "commandName": "run",
+      "commandLineArgs": "c:\\src\\xunit-performance-ericeil\\samples\\SimplePerfTests\\bin\\Release\\SimplePerfTests.dll -runner c:\\src\\xunit-performance-ericeil\\packages\\xunit.runner.console.2.1.0-rc1-build3168\\tools\\xunit.console.exe -runid test",
```
pharring: Does this need to contain user-specific folder structure?

dnx/xunit.performance.run.dnx/project.json
```diff
@@ -0,0 +1,29 @@
+{
+  "version": "1.0.0-*",
+  "description": "xunit.performance.run.dnx Console Application",
+  "authors": [ "ericeil" ],
+  "tags": [ "" ],
+  "projectUrl": "",
+  "licenseUrl": "",
+
+  "dependencies": {
+    "Microsoft.DotNet.xunit.performance.run.core": "1.0.0-alpha-build0015",
```
pharring: Consider whether this version number should be the "99.99.99-dev" version that we use elsewhere.
My guess is yes. For the nighly build, we used to have a task in the custom .msbuild file that replaced all occurrences of that version in all project.json files. It's disabled or commented out at the moment but we could add it back.

samples/DNXLibrary/project.json
```diff
@@ -9,13 +9,12 @@
     "dependencies": {
       "xunit": "2.1.0-rc1-build3168",
       "xunit.runner.dnx": "2.1.0-beta5-build169",
-      "Microsoft.DotNet.xunit.performance": "99.99.99-dev"
+      "Microsoft.DotNet.xunit.performance": "1.0.0-alpha-build0015"
```
pharring: As above, this could/should be the 99.99.99-dev version with automatic replacement in the .MSBuild file.
The idea is that, while developing locally that's the version you'll use. When it comes to release the NuGet packages, then we'll use the real semantic version.
Note that this is the same technique that xunit itself uses.

PR #80, opened 2015-09-28
Move to xunit 2.1.0 final

ericeil:
- Move to xunit 2.1.0 final
- Move to dnx.xunit 2.1.0-beta5a
- Fix a couple of build/project issues


ericeil:
@pharring, @mellinoe, @ianhays

pharring:
:+1:

pharring:
.. although take note of the Jenkins failure in Debug. Is that due to the change in the .MSBuild file where you build packages even for Debug builds?

ericeil:
Yes, the build issue was legit.  Added a commit that should fix this.

ericeil:
@dotnet-bot test this please

ericeil:
@mellinoe, @ianhays, are you ok with moving to the final xunit 2.1.0 release?

ianhays:
I haven't investigated what the changes that Release brings yet, but I'd imagine we want to move to eventually. With that said, I'll likely be too busy finding/squashing perf bugs to make the upgrade until ~2 weeks from now at least. I'll let Eric chime in on what he thinks though.

mellinoe:
I don't see any reason why we couldn't upgrade. Assuming there haven't been any assemblies renamed or breaking changes in the xUnit API, it should be fairly straightforward. I wouldn't be able to do it any time soon, but I don't think there's anything blocking us from moving to it when we need the latest perf tools

ericeil:
There are no breaking changes, as far as I can tell.  There are actually very, very few changes since RC1.
Thanks all, I'll go ahead with this.

PR #97, opened 2015-10-29
Support for MinIterations

### LINKED ISSUE #91: Minimum number of iterations?
Think something like this may have value? I've noticed from my data that many tests will only run one or two iterations and the data will be skewed oddly. An example:

```
      <test name="System.Diagnostics.Tests.Perf_Process.Kill" type="System.Diagnostics.Tests.Perf_Process" method="Kill" time="11.4533988" result="Pass">
        <performance runid="System.Diagnostics.Process.Tests.dll-WindowsCore" etl="D:\git\corefx\bin\tests\Windows_NT.AnyCPU.Release\System.Diagnostics.Process.Tests\dnxcore50\System.Diagnostics.Process.Tests.dll-WindowsCore.etl">
          <metrics>
            <Duration displayName="Duration" unit="msec" />
          </metrics>
          <iterations>
            <iteration index="0" Duration="248.70042802554235" />
            <iteration index="1" Duration="93.757350318643148" />
          </iterations>
        </performance>
      </test>
```

It seems like we would want to run this test a few more times to get a better idea of the expected duration.

ericeil:
Execute for a minimum number of iterations before stopping due to any time limits.  This way longer-running tests will still produce enough data for analysis.

Fixes #91.  @ianhays

### DIFF
--- a/samples/SimplePerfTests/Calibration.cs
+++ b/samples/SimplePerfTests/Calibration.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Xunit.Performance;
+using System;
 using System.Threading;
 using Xunit;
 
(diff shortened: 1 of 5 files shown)


ericeil:
@dotnet-bot test this please

PR #103, opened 2015-11-12
Add MeasureGCAllocationsAttribute

ericeil:
This adds a metric to track the size of allocations made on the GC heap.
Also, moves the GC trigger from being per-iteration to per-testcase.  This gives more accurate sampling of GC counts and allocations.  This may result in noisier durations, but on average they may be a more accurate representation of real performance; we'll need to gain some experience with this.

@sokket

### DIFF
--- a/samples/SimplePerfTests/StringThroughput.cs
+++ b/samples/SimplePerfTests/StringThroughput.cs
@@ -9,6 +9,7 @@ using System.Runtime.CompilerServices;
 using Xunit;
 
 [MeasureGCCounts]
+[MeasureGCAllocations]
 public static class StringThroughput
 {
     #region helpers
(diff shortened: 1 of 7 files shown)


sokket:
Awesome, thanks for doing this! What would the usage look like to gather allocation information during a test case?

ericeil:
Put `[MeasureGCAllocations]` on a test method, class, or assembly, and allocation info should appear in the results.

sokket:
Is there a way to Assert on that information during the test case? For instance, if we know the allocations should be below a certain threshold, can we ```Assert(allocatedBytes <= [some number])```?

ericeil:
No, that would need to be done in the analysis afterwards.  The allocation info comes from ETW, and isn't available synchronously at runtime.

PR #106, opened 2015-11-24
fix name of Duration metric in the analyzer

swgillespie:
The analysis executable assumed that the name of the test duration metric was `"duration"`, with a lowercase D, which caused the MetricValues dictionary to throw since the duration metric is called `"Duration"`, with an uppercase D. This PR lifts that value into a constant and fixes all occurrences of referring to the duration metric in this dictionary with a lowercase D.


ericeil:
Thanks!

PR #112, opened 2015-12-07
Expose the metrics assembly as a NuGet package

swgillespie:
Right now, it's currently not possible to define custom metrics without adding an assembly reference to the metrics assembly, which is itself impossible when compiling with `dnu` on ASP.NET 5 since it disallows non-NuGet references entirely.

I'm in the process of validating that this works correctly at runtime, so it's not quite ready to merge, but my custom metrics compile successfully on `dnu` targeting .NET 4.6 using this package. Since the metrics assembly has a dependency on TraceEvent, so does this package, which restricts its use to .NET 4.6.

### DIFF
--- a/src/xunit.performance.metrics.nuspec
+++ b/src/xunit.performance.metrics.nuspec
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.8.5">
+    <id>Microsoft.DotNet.xunit.performance.metrics</id>
+    <version>99.99.99-dev</version>
+    <title>xUnit Performance Metrics</title>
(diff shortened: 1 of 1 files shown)


ericeil:
Seems like a good idea to me.  If your testing goes well, let's go ahead and merge this.

swgillespie:
@ericeil I just did a successful benchmark run with custom metric collection, so it would appear to work! Since the windows runner NuGet package also provides the same metric assembly, this package may only be useful for compilation and be subsumed by the windows runner's assembly during test execution (which is fine, I guess - it's the same assembly either way).

ericeil:
Cool.  Thanks for doing this!

swgillespie:
No problem!

PR #114, opened 2016-01-21
Update xunit.performance.analysis to have a CSV option

davmason:
So we can generate files for submitting to the perf viewer


vancem:
LGTM

Issue #104, opened 2015-11-17
Ignore warmup iterations


Zoe-ms:
I tried the xunit-performance, and by default, it would run for 1000 times for every test, and seems the mean duration for a test would be calculated based on the 1000 durations. However, the first few iterations are doing warm-up, and better to be considered as exceptional value. 

For example: 
     `<iterations>`
           `<iteration index="0" Duration="0.90437474681425556" GCCount="0" />`
            `<iteration index="1" Duration="0.0021130251095655694" GCCount="0" />`
           `<iteration index="2" Duration="0.00090558218971636961" GCCount="0" />`
`...`

Thanks

ericeil:
The current convention is that iteration 0 should be treated as a "warmup" iteration for analysis purposes.  If we have a need for multiple warmup iterations, we can discuss ways of accommodating that.

luoyan0517:
I agree that at least iteration 0 should be treated as a "warmup" iteration, which is not implemented yet, right? As for if we need multiple warmup iterations, i can't say without any statistical data.

AndyAyersMS:
The analysis doesn't actually do this yet, right? I am seeing iteration 0 times included in the aggregated data....

vancem:
@dsgouda

brianrob:
This should be fixed now.

PR #115, opened 2016-02-26
Ignore iteration 0 as warmup in xunit.performance.analysis #104

luoyan0517:


ericeil:
Looks good to me.  @davmason should probably also have a look.

davmason:
Looks good to me too

PR #119, opened 2016-03-15
Fixing escaping for multiple commas in csv

dsgouda:
Fixed a bug where the escaping logic would not take into consideration strings with multiple commas in them
@davmason @brianrob

### DIFF
--- a/src/xunit.performance.analysis/Program.cs
+++ b/src/xunit.performance.analysis/Program.cs
@@ -286,14 +286,16 @@ namespace Microsoft.Xunit.Performance.Analysis
 
         private static string EscapeCsvString(string str)
         {
+            str.Trim('\"');
             // Escape the csv string
-            if(str.Contains("\""))
+            if (str.Contains("\""))
             {
-                str = "\"" + str.Replace("\"", "\"\"") + "\"";
+                str = str.Replace("\"", "\"\"");
             }
+            
             if (str.Contains(","))
             {
-                str = string.Format("\"{0}\"", str);
+                str = "\"\"" + str + "\"\"";
             }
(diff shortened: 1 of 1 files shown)


src/xunit.performance.analysis/Program.cs
```diff
@@ -287,14 +287,17 @@ private static void WriteTestResultsHtml(Dictionary<string, Dictionary<string, T
...
         {
             // Escape the csv string
-            if(str.Contains("\""))
+            if (str.Contains("\""))
             {
-                str = "\"" + str.Replace("\"", "\"\"") + "\"";
+                str = str.Replace("\"", "\"\"");
             }
-            if (str.Contains(","))
+            int numOfCommas = str.Split(',').Length - 1;
```
pharring: There's got to be a better way. String.Split is a memory hog.
Is the idea to add a pair of enclosing quotes for each comma?

```C#
using System.Linq;
int numOfCommas = str.Count(ch => ch == ',');
string enclosingQuotes = new string('\"', numOfCommas);
return enclosingQuotes + str + enclosingQuotes;
```

dsgouda: Nice! Will implement this. Yes, I am enclosing the string as many times as the number of commas.

davmason:
Why do you need to add a quote for every comma? Shouldn't one set of quotes be sufficient?

Also, it would be nice if you could rebase the multiple commits in to one before checking in.

dsgouda:
That is what it did earlier but that doesn't apparently work when we have strings with multiple commas.

davmason:
LGTM

PR #121, opened 2016-03-28
Switch to a fixed version of the dotnet cli

brianrob:
Fix the build by switching to a fixed version of the dotnet CLI.  Previously, we downloaded the latest build.


dsgouda:
:+1:  LGTM

PR #122, opened 2016-03-30
Add An Option to XUnit.Performance.Analysis to Dump Test Statistics

brianrob:
This option generates a CSV that provides per-test statistics for use in analyzing the effectiveness of the tests themselves.

### DIFF
--- a/src/xunit.performance.analysis/Program.cs
+++ b/src/xunit.performance.analysis/Program.cs
@@ -37,6 +37,7 @@ namespace Microsoft.Xunit.Performance.Analysis
             var xmlOutputPath = (string)null;
             var htmlOutputPath = (string)null;
             var csvOutputPath = (string)null;
+            var statsCsvOutputPath = (string)null;
 
             for (int i = 0; i < args.Length; i++)
             {
@@ -71,6 +72,11 @@ namespace Microsoft.Xunit.Performance.Analysis
                             if (++i >= args.Length)
                                 return Usage();
                             csvOutputPath = args[i];
+                            break;
+                        case "stats":
+                            if (++i >= args.Length)
+                                return Usage();
+                            statsCsvOutputPath = args[i];
                             break;
 
                         default:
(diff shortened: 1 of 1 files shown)


brianrob:
@davmason, can you please review this?

src/xunit.performance.analysis/Program.cs
```diff
@@ -318,6 +327,44 @@ private static void WriteTestResultsCSV(Dictionary<string, Dictionary<string, Te
...
+                        RunningStatistics durationStats = result.Stats[DurationMetricName];
+                        writer.WriteLine(
+                            "\"{0}\",\"{1}\",\"{2}\",\"{3}\",\"{4}\",\"{5}\",\"{6}\"",
+                            EscapeCsvString(result.TestName),
+                            durationStats.Count,
+                            durationStats.Minimum,
+                            durationStats.Maximum,
+                            durationStats.Mean,
+                            durationStats.StandardDeviation,
+                            GetMetricsString(result.Stats.Keys)
```
dsgouda: If this is a string, we might want to escape it

brianrob:
@dotnet-bot test this please

src/xunit.performance.analysis/Program.cs
brianrob: Which specific field are you thinking should be escaped?

dsgouda: "GetMetricsString(result.Stats.Keys)"  if I understand correctly it should be the metrics which ideally should not have any special characters but just to be on the safer side...

brianrob: Ah, yes... you're right.  Fixed.  Thanks for catching this.

davmason:
Looks good to me

brianrob:
@dotnet-bot test this please

brianrob:
Thanks all for the review.

PR #123, opened 2016-04-18
Build dotnet cli runner using rc2 CoreFX

dsgouda:
- Building dotnet cli runner using CoreFX since that would generate portable binaries
- rc3 CoreFX packages are not supported to be portable currently
- These published binaries will allow us to run performance runner on non-windows systems
- Building xunit.performance analysis for both cli (using project.json) and desktop (using csproj)
- Added conditional compile for code in xunit.performance.analysis since MathNet.Numerics package is not available for non-desktop platforms, cli version of xunit.performance.analysis can only convert the xmls to csv for benchview consumption
- Modifying BuildForPublication command to install dotnet cli using the recommended dotnet-install.ps1 script authored by the dotnet cli team, this would take care of automatic dotnet cli version updates and bug fixes too

@brthor @brianrob @lt72


src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
```diff
@@ -20,17 +20,18 @@
   },
 
   "frameworks": {
-    "dnxcore50": {
-      "imports": "portable-net45+win8",
+    "netstandard1.5": {
```
brthor: Why are you using net standard here? You have `emitEntryPoint: true` so I assume this is an application?

`netcoreapp1.0` is the official TfM for this.

src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
```diff
@@ -20,17 +20,18 @@
...
+      "imports": [
+        "dnxcore50",
+        "portable-net45+win8"
+      ],
       "dependencies": {
-        "NETStandard.Library": "1.0.0-rc3-23904",
+        "Microsoft.NETCore.App": {
+          "type": "platform",
+          "version": "1.0.0-rc2-*"
```
brthor: Although I gave you this version string for testing, I recommend you use "1.0.0-rc2-3*" for the string here as it will ensure you are only getting the newest versions of this package (the older ones will break)

src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
dsgouda: I copied it from here
https://github.com/dotnet/cli/blob/rel/1.0.0/TestAssets/TestProjects/PortableTests/PortableApp/project.json#L7
Will make the changes now

src/xunit.performance.analysis/Program.cs
```diff
@@ -41,7 +45,7 @@ private static int Main(string[] args)
 
             for (int i = 0; i < args.Length; i++)
             {
-                if (args[i].StartsWith("-") || args[i].StartsWith("/"))
```
lt72: are you removing this because it was interfering with some other parameters, or because other tools do not support the '/' notation?

src/xunit.performance.analysis/Program.cs
```diff
@@ -116,21 +120,25 @@ private static int Main(string[] args)
 
             var comparisonResults = DoComparisons(allComparisonIds, testResults);
 
+        #if !LINUX_BUILD
```
lt72: how about using the #if contruct only to stub out the WeiteXXX methods?  so we have less #if overall...

src/xunit.performance.analysis/Program.cs
```diff
@@ -192,10 +210,32 @@ private static List<TestResultComparison> DoComparisons(List<Tuple<string, strin
 
                 foreach (var metric in iteration.MetricValues)
                 {
+                #if !LINUX_BUILD
```
lt72: maybe it would be better to just have a different SummarizeTestResults method for Linux builds, rather than the same method with many #if inside

src/xunit.performance.analysis/Program.cs
```diff
@@ -311,22 +353,28 @@ private static string EscapeCsvString(string str)
 
         private static void WriteTestResultsCSV(Dictionary<string, Dictionary<string, TestResult>> testResults, string csvOutputPath)
         {
+        #if LINUX_BUILD
```
lt72: it seems the Linux approach here (using FileStream) could work for Windows too, so we get rid of one #if

src/xunit.performance.analysis/Program.cs
dsgouda: The Linux file paths start with /home and it ended up reading those as parameters, I could probably add a #if for Linux here but thought this would be more consistent

dsgouda:
@lt72 added the suggested changes, please take a look

dsgouda:
@dotnet-bot Test Release Build please

src/xunit.performance.analysis/Program.cs
lt72: understood, I think your solution is OK!

lt72:
👍

lt72:
Build error is about missing package for TraceEvent: 

```
This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is ..\..\packages\Microsoft.Diagnostics.Tracing.TraceEvent.1.0.39\build\Microsoft.Diagnostics.Tracing.TraceEvent.targets.
  10:2>Done executing task "Error" -- FAILED.
```

dsgouda:
@dotnet-bot Test Release please

dsgouda:
@dotnet-bot Test this please

PR #126, opened 2016-04-26
Building cli projects for netstandard1.3

dsgouda:
netcoreapp1.0 is not supported with dnxcore50 for certain projects in CoreFX, building for netstandard1.3 instead which is supported instead
Adding DotNetCliVersion.txt to control what version of dotnet cli we are using

@brianrob @lt72

PR #127, opened 2016-05-02
Lock dotnet version to unblock perf testing

dpodder:
CiBuild was downloading the 'latest' dotnet framework version, which broke our
ability to run dotnet performance tests. This PR locks in the correct version to
unblock perf testing.


dpodder:
@brianrob PTAL

brianrob:
To add a bit of context, the issue is that the CLI-built components use a * for the version of the shared framework, and the CLI downloader script always downloads Latest.  However, these two versions don't always match - thus when we try to run the CLI components, corehost complains that it doesn't have the targeted version of the shared framework installed.

The solution as @dpodder explains above is to specify the version directly.

@dsgouda, do you have an issue tracking bringing back the CLI version file?  If so, let's link to it here.

brianrob:
LGTM.

@dpodder, I will merge this as soon as the CI is green.

dsgouda:
@brianrob I have opened a PR to address the issue you pointed out here https://github.com/Microsoft/xunit-performance/pull/126

PR #128, opened 2016-05-06
Trimming cli dependencies to NETStandard.Library

dsgouda:
Building executable libraries using NETStandard library since we cannot use Microsoft NetCoreApp dependencies for CoreFX

@lt72 @ericstj

### DIFF
--- a/NightlyBuild.cmd
+++ b/NightlyBuild.cmd
@@ -19,8 +19,8 @@ bump the build number on BuildSemanticVersion below.
 :main
 setlocal
 
-set BuildAssemblyVersion=1.0.0.32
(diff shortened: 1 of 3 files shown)


ericstj:
Looks ok as a temporary solution let's make sure to get bugs on CLI to refine this app-model.  One thing I thought of that seems more natural than this would be to reference NETCore.App as `"suppressParent": "all"` but that seems to be ignored by `dotnet pack`.

dsgouda:
Thanks @ericstj  I will open a separate issue on dotnet cli to report the lack of support for "suppressParent" flag

dsgouda:
@dotnet-bot Test Release Build please

PR #130, opened 2016-05-09
Fix build breaks

brianrob:
Fix build breaks:

1. Windows x86 (or x64 under SysWOW shell) - RID missing.
2. Ubuntu - RID missing.
3. NuGet.config relative path to LocalPackages not parsed properly by CLI on Ubuntu.


brianrob:
@dsgouda, PTAL

src/cli/Microsoft.DotNet.xunit.performance.analysis.cli/project.json
```diff
@@ -33,6 +33,8 @@
     }
   },
   "runtimes": {
-    "win7-x64": {}
+    "win7-x86": {},
```
dsgouda: I still have reservations about explicitly adding "win7-x86" as a runtime. Do we have x86 machines for CI? I think we should ask someone from the dotnet cli team to verify this for us. 
@brianrob thoughts?

dsgouda: Had a chat with dotnet cli folks, ideally for libraries we don't need to specify the rids but since we need an entryPoint for the runner/analyzer and hence we need to specify the rids. I have filed an issue here to track this https://github.com/dotnet/cli/issues/2913
My other idea was to not specify rids at all and use --infer-runtimes when restoring, but it seems dotnet cli will not be supporting it moving forward.

dsgouda:
LGTM

PR #139, opened 2016-05-11
Plumb build configuration to BuildCliComponents.cmd

brianrob:
Prior to this change, BuildCliComponents.cmd always build Release.  Now by default it builds debug, but respects command line input to specify the build configuration.

The build configuration is also specified via NightlyBuild.cmd (which always builds release) and CiBuild.cmd (which builds both debug and release based on input).

### DIFF
--- a/BuildCliComponents.cmd
+++ b/BuildCliComponents.cmd
@@ -13,6 +13,9 @@ tools directory prior to build.
 :main
 setlocal
 
+set BuildConfiguration=%1%
+if "%BuildConfiguration%"=="" set BuildConfiguration=Debug
+
 set OutputDirectory=%~dp0LocalPackages
 set DotNet=%~dp0\tools\bin\dotnet.exe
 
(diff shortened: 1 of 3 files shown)


brianrob:
@dotnet-bot test this please

brianrob:
@dsgouda, PTAL

BuildCliComponents.cmd
```diff
@@ -40,13 +43,13 @@ if NOT exist "%DotNet%" (
...
-call %DotNet% pack -c Release -o %OutputDirectory%
+call %DotNet% build -c %BuildConfiguration%
+call %DotNet% pack -c %BuildConfiguration% -o %OutputDirectory%
 popd
 pushd %~dp0src\cli\Microsoft.DotNet.xunit.performance.analysis.cli
 call %DotNet% restore
-call %DotNet% build -c Release
-call %DotNet% pack -c Release -o %OutputDirectory%
+call %DotNet% build -c %BuildConfiguration%
+call %DotNet% pack -c %BuildConfiguration% -o %OutputDirectory%
```
dsgouda: Not sure if we want to pack a Debug version, maybe we can pack only if it is a Release build

brianrob: What is the concern around this?

dsgouda: Its probably nothing but if we forget to specify the config when publishing we might publish the debug version of nupkg by default

dsgouda:
LGTM

BuildCliComponents.cmd
brianrob: Gotcha.  I do want us to be able to build debug nuget packages for when we need them for testing.  However, you reminded me of another issue that we do need to solve, so I've filed it at #141.

PR #143, opened 2016-05-19
Packaging cli version number with binaries

dsgouda:
-Packaging the file indicating version number of cli used to build the libraries so we can use the same version to execute them

@brianrob PTAL


brianrob:
LTGM

PR #144, opened 2016-05-19
Apply existing comparison analysis to all observed metrics. 

lt72:
Refactored xunit performance analysis tool to show existing, minimal statistical analysis for all observed metrics. 
Created extensible writers for all currently used file formats.
Small addition to .gitignore

Code that writes to .csv, .xml and .http file formats has moved to the writers. The .html writer includes a small header to list all collected metrics and data points available for each metric. Observed metrics are collected at when ETW logs are parsed, and statistics are computed for all observed metrics, if data points are available.

### DIFF
--- a/src/xunit.performance.analysis/Program.cs
+++ b/src/xunit.performance.analysis/Program.cs
@@ -3,29 +3,23 @@
 
 #if !LINUX_BUILD
 using MathNet.Numerics.Statistics;
-using Microsoft.Diagnostics.Tracing;
-using Microsoft.Diagnostics.Tracing.Parsers;
+using System.Runtime.Serialization;
 #endif
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Xml;
 using System.Xml.Linq;
-#if !LINUX_BUILD
-using System.Runtime.Serialization;
-#endif
 
 namespace Microsoft.Xunit.Performance.Analysis
 {
     internal class Program
     {
-        private const double ErrorConfidence = 0.95; // TODO: make configurable
-
         /// <summary>
-        /// The name of the Duration metric, as provided by the XML.
+        /// The set of metrics (name of the metric) that appear in the .etl files
         /// </summary>
-        private const string DurationMetricName = "Duration";
 
         private static int Usage()
         {
@@ -94,8 +88,11 @@ namespace Microsoft.Xunit.Performance.Analysis
                     {
                         if (file.EndsWith(".xml", StringComparison.OrdinalIgnoreCase))
                         {
-                            foundFile = true;
-                            xmlPaths.Add(file);
+                            if (file != xmlOutputPath)
+                            {
+                                foundFile = true;
+                                xmlPaths.Add(file);
+                            }
                         }
                         else
                         {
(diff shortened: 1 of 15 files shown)


lt72:
@brianrob: PTAL

src/xunit.performance.analysis/Program.cs
```diff
@@ -14,27 +14,25 @@
...
+        
         private static int Usage()
         {
             Console.Error.WriteLine(
                 "usage: xunit.performance.analysis <xmlPaths> [-compare \"baselineRunId\" \"comparisonRunId\"]  [-xml <output.xml>] [-html <output.html>]");
             return 1;
         }
 
-        private static int Main(string[] args)
+        private static int Main( string[ ] args )
```
pharring: You appear to be using custom spacing options. I suggest resetting to the Visual Studio defaults for C#.

lt72: yes, that fell through before I reset.  Is there a coding convention file or just default VS2015 settings?

src/xunit.performance.analysis/Program.cs
```diff
@@ -43,256 +41,179 @@ private static int Main(string[] args)
...
+                        var baselineSumSquared = baselineSum * baselineSum;
+                        var baselineSumOfSquares = baselineTest.Iterations.Sum(iteration => iteration.MetricValues[metricName] * iteration.MetricValues[metricName]);
+                        
+                        var comparisonSum = comparisonTest.Iterations.Sum(iteration => iteration.MetricValues[metricName]);
+                        var comparisonSumSquared = comparisonSum * comparisonSum;
+                        var comparisonSumOfSquares = comparisonTest.Iterations.Sum(iteration => iteration.MetricValues[metricName] * iteration.MetricValues[metricName]);
+
+                        var stdErrorDiff = Math.Sqrt((baselineSumOfSquares - (baselineSumSquared / baselineCount) + comparisonSumOfSquares - (comparisonSumSquared / comparisonCount)) * (1.0 / baselineCount + 1.0 / comparisonCount) / (baselineCount + comparisonCount - 1));
+                        var interval = stdErrorDiff * MathNet.Numerics.ExcelFunctions.TInv(1.0 - Properties.ErrorConfidence, baselineCount + comparisonCount - 2);
```
AndyAyersMS: We're assuming here that the data here has a normal distribution. This is unlikely for performance data, especially given the generally small default sample size (10).

I realize you're just reworking preexisting code, but we should consider using a non-parametric estimation technique for metrics on populations, and some variant of hypothesis-based testing for determining if some metric is significantly different between populations.

src/xunit.performance.analysis/xunit.performance.analysis.csproj
```diff
@@ -65,6 +74,7 @@
     <None Include="App.config" />
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup />
```
brianrob: Superfluous <ItemGroup /> tag?

src/xunit.performance.analysis/Writers/XmlResultsWriter.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                var comparisonResults = ComparisonResults[metricName];
+                
+                foreach(var comparison in comparisonResults)
+                {
+                    var comparisonElem = new XElement("comparison", new XAttribute("test", comparison.TestName), new XAttribute("baselineId", comparison.BaselineResult.RunId), new XAttribute("comparisonId", comparison.ComparisonResult.RunId));
+                    resultElem.Add( comparisonElem );
+
+                    comparisonElem.Add(
+                        new XElement( "duration",
```
brianrob: I think you want to replace "duration" with the metric name.

src/xunit.performance.analysis/Program.cs
```diff
@@ -454,6 +310,23 @@ private static IEnumerable<TestIterationResult> ParseOneXmlFile(string path)
...
+                {
+                    foreach(var metric in metrics.Elements())
+                    {
+                        var metricName = metric.Name.LocalName;
+                        var unit       = metric.Attribute("unit").Value;
+
+                        //
+                        // Populate the set of all collected metrics 
+                        //
+                        if(Properties.AllMetrics.ContainsKey( metricName ) == false)
```
brianrob: Nit: Replace == false with !

brianrob:
@dsgouda, you should look at this as it touches csv generation.

src/xunit.performance.analysis/Program.cs
brianrob: Agreed that we should be considering better statistics over the data.

brianrob:
@lt72, overall, looks good to me modulo the few comments.  Please do also fix up the spacing as @pharring suggests - looks like a big chunk of the diff is just whitespace.

src/xunit.performance.analysis/Program.cs
lt72: Absolutely agree, I will put this work on the books for early next sprint. 

---
In reply to: [63909764](https://github.com/Microsoft/xunit-performance/pull/144#discussion_r63909764) [](ancestors = 63909764)

src/xunit.performance.analysis/Program.cs
lt72: Should I assume that a coding convention? '== false' is more readable for me, but consistency is a value, if indeed we have guidance at that level. Is there any enforcing tool we are using in the team?

---
In reply to: [63916980](https://github.com/Microsoft/xunit-performance/pull/144#discussion_r63916980) [](ancestors = 63916980)

brianrob: Yes, this is just based on the existing conventions in the code.  There isn't a tool that we are using to enforce this - same goes for other repos like coreclr and corefx.

lt72:
@dotnet-bot Test this please

lt72:

lt72:

lt72:
Fixes #146

PR #153, opened 2016-06-01
Update performance runners to restore for netstandard1.3

karajas:
This change updates the perf libraries to restore for netstandard1.3.

This involves them consuming the repackaged xunit libraries from buildtools and is dependent on PR https://github.com/dotnet/buildtools/pull/770 . I have made the Microsoft.xunit.extensibility.execution.netcore dependency versionless so that it will pick up the most recent copy(we shouldn't any updates to this package ever - this is simply a workaround until xunit ships with rc2 support).

/cc @dsgouda @brianrob @ericstj

### DIFF
--- a/src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
+++ b/src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
@@ -20,7 +20,8 @@
     "Microsoft.DotNet.xunit.performance.run.core": "99.99.99-dev",
     "Microsoft.NETCore.DotNetHostPolicy":"1.0.1-rc2-002357-00",
     "NETStandard.Library": "1.5.0-rc2-23931",
-    "xunit.extensibility.execution": "2.1.0"
+    "Microsoft.xunit.extensibility.execution.netcore": "1.0.0-prerelease-00508-01",
+    "System.Diagnostics.Process": "4.1.0-rc2-24103",
   },
   "__comment": "netcoreapp1.0 import and the runtime below is a hack, we need CLI to allow us to build issue reference: https://github.com/dotnet/cli/issues/2913",
   "frameworks": {
--- a/NightlyBuild.cmd
+++ b/NightlyBuild.cmd
@@ -19,8 +19,8 @@ bump the build number on BuildSemanticVersion below.
 :main
 setlocal
 
-set BuildAssemblyVersion=1.0.0.35
-set BuildSemanticVersion=1.0.0-alpha-build0035
+set BuildAssemblyVersion=1.0.0.36
+set BuildSemanticVersion=1.0.0-alpha-build0036
 set OutputDirectory=%~dp0LocalPackages
 set DotNet=%~dp0\tools\cli\bin\dotnet
 
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <add key="Locally built packages" value="LocalPackages" />
+    <add key="dotnet-buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
--- a/src/xunit.performance.nuspec
+++ b/src/xunit.performance.nuspec
@@ -39,6 +39,25 @@
         <dependency id="xunit.extensibility.core" version="2.1.0" />
         <dependency id="xunit.extensibility.execution" version="2.1.0" />
       </group>
+      <group targetFramework="netstandard1.3">
+        <dependency id="System.Collections" version="4.0.10" />
+        <dependency id="System.Diagnostics.Debug" version="4.0.10" />
+        <dependency id="System.Diagnostics.Tracing" version="4.0.20" />
+        <dependency id="System.Globalization" version="4.0.10" />
+        <dependency id="System.IO" version="4.0.10" />
+        <dependency id="System.IO.FileSystem" version="4.0.0" />
+        <dependency id="System.IO.FileSystem.Primitives" version="4.0.0" />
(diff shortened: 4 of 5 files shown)


karajas:
One thing I am still unsure about is how this may affect other repos consuming these perf runners, which may not have updated to netstandard. A workaround for this could be to repackage this runner into a netcore version similar to buildtools.

brianrob:
@karajas, can you share some information on why this change is needed?  Thanks.

karajas:
Sure, we are trying to update corefx to start restoring tests for netstandard so that all of them can compile against the netstandard contracts. The idea is to make all the tests portable for that netstandard version and then deploy their runtime assets according to the target framework(net46, netcoreapp1.0, netcore50, etc.). Since the corefx test-runtime which is restored for every project has a dependency on the perf runner libraries (which currently restore for the **dotnet** tfm) , this change would repackage those deps to restore the test-runtime for netstandard.

A thing to note is that this is not a true update to netstandard but a repackaging as I believe the perf libs haven't been updated to use the shipped rc2 packages.

ericstj:
> One thing I am still unsure about is how this may affect other repos consuming these perf runners, which may not have updated to netstandard. A workaround for this could be to repackage this runner into a netcore version similar to buildtools.

'netstandard` monikers are compatible with all old profiles so this should just work.

brianrob:
Thanks @karajas.  As you point out, I suspect that this change will have an impact on other repos that don't use netstandard.  We've had a huge amount of difficulty keeping the infrastructure up, and given that we're not far from RTM, I don't see us being able to take this prior to RTM.  What happens if we don't do this?

ericstj:
If you're worried about breaking existing repos you can have leave both sections in the packages.

brianrob:
@ericstj, yup, worried about breaking existing repos.

I'll let @lt72 comment here as well, but the bar here before RTM is extremely high as every time we make some kind of change to the perf infra, things break and it takes 1-2 days to resolve.

If you believe that these changes are compatible, then min-bar testing is validation of the new packages against the end-to-end pipeline in Helix against CoreFX on Windows and Linux and CoreCLR on Windows (Linux is still coming up).

lt72:
@karajas: am I wrong or this change requires changes in buildtools for CloudTest.targets (per your email). Cross repo changes have been extremely problematic and we need to validate extensively if that is the case. The bar Brian mentions is correct.

src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
```diff
@@ -20,7 +20,8 @@
     "Microsoft.DotNet.xunit.performance.run.core": "99.99.99-dev",
     "Microsoft.NETCore.DotNetHostPolicy":"1.0.1-rc2-002357-00",
     "NETStandard.Library": "1.5.0-rc2-23931",
-    "xunit.extensibility.execution": "2.1.0"
+    "Microsoft.xunit.extensibility.execution.netcore": "1.0.0-prerelease",
```
ericstj: This package needs a more specific version I believe.  I suspect you will update after you get a build-tools build?

brianrob: @ericstj, are you just saying that Microsoft.xunit.extensibility.execution.netcore doesn't have a full version number listed?

karajas: I was hoping to let this be a 1.0.0 version if possible. We wouldn't need to update this until xunit gets updated.

karajas:
This change requires https://github.com/dotnet/buildtools/pull/770 to be approved. But the fixes this would involve are trivial - https://github.com/karajas/buildtools/commit/77975b0899d318d46b089028af08ec6fcaded462 . This is because perf uses the same targets for CoreCLR and CoreFX. I've tested these changes for Windows and Linux in CoreFX. I'm able to upload and execute the runner on CoreCLR.

src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
brianrob: Why is it that we need to replace xunit.extensibility.execution?  I get worried about replacing external dependencies with ones that we create.

brianrob:
@dotnet-bot Windows / Ubuntu Debug Build

brianrob:

src/cli/Microsoft.DotNet.xunit.performance.runner.cli/project.json
karajas: This was a repackaged version of the same fixed version of the dependency so that we are also able to restore for netstandard1.3. 

From https://github.com/karajas/buildtools/blob/eef26e8e6e46c4b11b721ce84c11bdd02fcc9207/src/nuget/Microsoft.xunit.extensibility.execution.netcore.nuspec ,
<file src="Microsoft.xunit.extensibility.execution.netcore\xunit.execution.dotnet.dll" target="lib/netstandard1.0" />

karajas:

karajas:
@brianrob PTAL. xunit.extensibility.execution was an external dep that supported only older tfms and this repackaged version of the same fixed version of the dependency is so that we are also able to restore for netstandard1.3.

From https://github.com/karajas/buildtools/blob/eef26e8e6e46c4b11b721ce84c11bdd02fcc9207/src/nuget/Microsoft.xunit.extensibility.execution.netcore.nuspec , the file included was "Microsoft.xunit.extensibility.execution.netcore\xunit.execution.dotnet.dll" for target="lib/netstandard1.0"

brianrob:
Thanks @karajas.  Is there any reason that we can't just get the official packages updated?

dsgouda:
I don't see anything that should block, LGTM subject to @brianrob 's final assessment

karajas:
@brianrob As far as I know, xunit has a build of rc2 that isn't stable enough to upgrade to across all of our repos. I think they are waiting on rtm. I think @ericstj can comment on this.

ericstj:
> Is there any reason that we can't just get the official packages updated?

There is an issue tracking that: https://github.com/xunit/xunit/issues/872

Until that point we will redistribute them in our internal infrastructure packages to avoid the import.

brianrob:
Ok, thanks for clarifying.

brianrob:
LGTM

PR #154, opened 2016-06-02
Fixing cli runner dependencies

dsgouda:
Revving System.Diagnostics.Process dependencies in xunit performance since the older version indirectly references a windows-only dll and cannot be run on linux
This fixes current perf runs on linux

@brianrob @lt72


brianrob:
If possible, let's use latest version number that's already published so that we don't have to update BuildTools and all of the downstream repos.  Otherwise LGTM.

dsgouda:
@dotnet-bot Test Windows / Ubuntu Debug Build please

PR #155, opened 2016-06-09
Update nuget reference

davmason:
We had a hard link on an old version of nuget that didn't work with the netstandard profile when creating nuget packages.


brianrob:
LGTM

PR #156, opened 2016-06-10
Remove xunit dependency from nuspec for netstandard and use the new buildtools package

karajas:
This is to remove the xunit dependencies preventing restore for netstandard. This will only affect the netstandard section of the nuspec.

/cc @dsgouda @brianrob

### DIFF
--- a/src/xunit.performance.nuspec
+++ b/src/xunit.performance.nuspec
@@ -54,9 +54,7 @@
         <dependency id="System.Text.Encoding" version="4.0.10" />
         <dependency id="System.Threading" version="4.0.10" />
         <dependency id="System.Threading.Tasks" version="4.0.10" />
-        <dependency id="xunit.abstractions" version="2.0.0" />
-        <dependency id="xunit.extensibility.core" version="2.1.0" />
-        <dependency id="xunit.extensibility.execution" version="2.1.0" />
+        <dependency id="Microsoft.DotNet.BuildTools.TestSuite" version="1.0.0-prerelease-508-01" />
       </group>
     </dependencies>
   </metadata>
--- a/NightlyBuild.cmd
+++ b/NightlyBuild.cmd
@@ -19,8 +19,8 @@ bump the build number on BuildSemanticVersion below.
 :main
 setlocal
 
-set BuildAssemblyVersion=1.0.0.36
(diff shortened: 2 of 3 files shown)


src/xunit.performance.nuspec
```diff
@@ -54,9 +54,7 @@
         <dependency id="System.Text.Encoding" version="4.0.10" />
         <dependency id="System.Threading" version="4.0.10" />
         <dependency id="System.Threading.Tasks" version="4.0.10" />
-        <dependency id="xunit.abstractions" version="2.0.0" />
-        <dependency id="xunit.extensibility.core" version="2.1.0" />
-        <dependency id="xunit.extensibility.execution" version="2.1.0" />
+        <dependency id="Microsoft.DotNet.BuildTools.TestSuite" version="1.0.0-prerelease-508-01" />
```
brianrob: Can you provide details on what this package is?  Is it a re-packaged version of the xunit dependencies?

karajas: Sure. Yes these are the repackaged xunit dependencies, containing https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.BuildTools.TestSuite/project.json 
https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.DotNet.BuildTools.TestSuite

brianrob: Thanks.  It occurs to me that we also need to make sure that this doesn't break the perf tooling in the CLI repo - things are slightly different there at this time.

Can you work with @dpodder to make sure that the CLI perf testing scripts work with these changes?

brianrob: To add a bit of context - I get concerned when we have to re-package because it means that there is a chance that dependencies change in some fashion (libraries, tfm, etc.)  CLI is different because we run under dotnet test.

karajas: Sure. Although again this shouldn't affect any other runs as it is adding an additional tfm to search for in the libs. From the cli project.json (https://github.com/dotnet/cli/blob/rel/1.0.0/test/Performance/project.json#L26 ), looks like it will only bring in the dotnet tfm dependencies.

brianrob: Gotcha, and thanks for checking on this.

brianrob:
LGTM

PR #164, opened 2016-07-25
Adding multi metric support to xunit perf analysis

dsgouda:
Enabling generation of multi-metric csvs which can then be displayed on benchview.
Generating one csv file per metric as per benchview specs
fixing some dependencies in nuspec files
@lt72 

P.S.: There is some testing pending however this does generate the appropriate csv files from test results xml.
This also needs fixing of the version number


lt72:
:+1:

PR #167, opened 2016-08-12
Fix bug in readme

robbert229:
In the read me references are made to `iteration.InnerIterationsCount`. This doesn't exist and at somepoint was renamed to `Benchmark.InnerIterationCount`


brianrob:
@robbert229, thanks very much for submitting a fix for this!  You're right - it was renamed prior to check-in.

PR #169, opened 2016-08-20
Add `BetterDirection` property to `PerformanceMetricInfo`

dpodder:


dpodder:
@brianrob @dsgouda Could you take a look at this PR for feedback, and maybe mention anyone else who would be a good person to review? The idea is to add a new property to a metric, "better", which is either ascending or descending depending on the desired direction of improvement (useful for, e.g., frames/second or benchmark score... cases where higher is better).

brianrob:
@dpodder, I need to think about this one just a bit.  Originally, the goal was to have all metrics have a single "better direction" - the lower the value, the better the result.

I did a bunch of thinking on this a while back, but I can't remember where I ended up.  Let me try to recall this, and I'll get back to you.

dpodder:
Thanks, @brianrob; feel free to reach out to discuss in more detail. If we are comfortable with locking in xunit-perf to always have `better == descending`, then this change is certainly not necessary; I originally just wanted to explore what it would take to implement it and found it relatively straightforward to do so, hence this PR.

brianrob:
@dpodder, I think I'd like to leave this as is, but I'm open to re-considering when we have a case for using this.  Thoughts?

dpodder:
Thanks @brianrob, I agree as this has not been an issue since August. I'll close this PR for now, particularly since this change was built on top of the legacy runner; we would need to reimplement it on top of the new API anyway.

PR #171, opened 2016-08-30
Fix InstructionsRetired Reports (Missing Data)

brianrob:
The instructions retired data was being thrown away because there were only 2 values across all iterations.  Our outlier analysis threw them out.

This change ensures that we don't throw away more than 5% of data - if we feel we need to then we don't throw any away.

@dsgouda, PTAL

### DIFF
--- a/NightlyBuild.cmd
+++ b/NightlyBuild.cmd
@@ -19,8 +19,8 @@ bump the build number on BuildSemanticVersion below.
 :main
 setlocal
 
-set BuildAssemblyVersion=1.0.0.40
-set BuildSemanticVersion=1.0.0-alpha-build0040
+set BuildAssemblyVersion=1.0.0.41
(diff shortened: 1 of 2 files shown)


dsgouda:
If you are planning to publish a new package version you might want to update https://github.com/Microsoft/xunit-performance/blob/master/NightlyBuild.cmd#L22 and https://github.com/Microsoft/xunit-performance/blob/master/NightlyBuild.cmd#L23

dsgouda:
I'm assuming the changes have been tested 
LGTM :+1:

brianrob:
@dsgouda, good point, I've updated NightlyBuild.cmd.  We can publish a new package version, but we don't need to modify CoreCLR or CoreFX, because they don't depend on the change (e.g. they haven't hit this bug).

PR #173, opened 2016-10-05
Linux: Fixing exception.

jorive:
"Invalid URI: The format of the URI could not be determined." exception was caused by passing a relative path to load the xml.


brianrob:
@jorive, is this still relevant?

PR #179, opened 2016-11-23
Implementing IDisposable on the csv classes.

jorive:


jorive:
@dpodder ping

src/xunit.performance.api/Table/CSVFile.cs
```diff
@@ -1,79 +1,84 @@
...
+
+            _titlePositions = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+            _lines = new List<string[]>();
 
             // Get the title line
-            string line = this._Reader.ReadLine();
+            var line = _reader.ReadLine();
             if (line == null)
-            {
                 throw new InvalidDataException("No lines in CSV file.");
```
dpodder: Don't we need to dispose both `_stream` and `_reader` here?

jorive: @dpodder good catch! I'll fix it!

PR #182, opened 2017-01-12
Add back strong name signing

DrewScoggins:
When we made the transition to use dotnet cli to build the project we
stopped strong name signing the resulting binaries.  This caused a
problem when we try and use these binaries to build our perf tests in
the CoreFX repo.  To fix this problem we just use the
GlobablAssemblyInfo file to apply the key to all of the projects.  We
also neede to update the AssemblyInfo of the execution project because
it allows the api project to view its internal state and since the
binaries are now strong name signed we need to include the public key.

### DIFF
--- a/src/common/GlobalAssemblyInfo.cs
+++ b/src/common/GlobalAssemblyInfo.cs
@@ -3,8 +3,5 @@
 
 using System.Reflection;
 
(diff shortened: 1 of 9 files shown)


DrewScoggins:
@dpodder @jorive  PTAL

PR #183, opened 2017-02-07
Adding ETW to xunit-performance API

jorive:
1. This work is based of the .NET Core build of the Microsoft.Diagnostics.Tracing.TraceEvent library.
2. Changed the statistics output to be markdown table (this adds dependency on MarkdownLog library build that was built for .NET Core).
3. Cleanup and refactor the build scripts.
4. Cleanup and refactor the csproj.
5. Added a Kernel32 class that PInvoke methods to determine the Windows version (current .NET API is only available on 1.7 which would not support .NET Core 1.0).

This is an example of the markdown text output by the harness:
 Test Name                                   | Metric   | Iterations | AVERAGE | STDEV.S |   MIN |   MAX
:------------------------------------------- |:-------- |:----------:| -------:| -------:| -----:| -----:
 simpleharness.Program+Type_1.TestBenchmark  | Duration |    101     |   1.686 |   0.248 | 1.493 | 2.981
 simpleharness.Program+Type_2.TestBenchmark1 | Duration |    101     |   1.651 |   0.188 | 1.459 | 2.499
 simpleharness.Program+Type_2.TestBenchmark2 | Duration |    101     |   1.589 |   0.123 | 1.468 | 2.038


brianrob:
@dotnet-bot test ci please

BuildForPublication.cmd
```diff
@@ -3,12 +3,21 @@ goto :main
...
 
-set PackageVersion=1.0.0-alpha-build0043
-set VersionSuffix=build0043
+  echo/==================
+  echo/ Building version %PackageVersion% NuGet packages.
+  echo/==================
 
-echo Building version %PackageVersion% NuGet packages.
+  set LocalDotNet_ToolsDir=%~dp0tools
+  if exist "%LocalDotNet_ToolsDir%" rmdir /s /q "%LocalDotNet_ToolsDir%"
```
brianrob: We should not delete the tools directory on every run.  Ideally you can build without internet access as long as you have already downloaded the toolset and nuget dependencies.

build.cmd
```diff
@@ -1,108 +1,220 @@
...
+  powershell -NoProfile -ExecutionPolicy unrestricted -Command "Invoke-WebRequest -Uri '%DotNet_Installer_Url%' -OutFile '%DotNet_Path%\dotnet-install.ps1'"
+  echo Executing dotnet installer script %DotNet_Path%\dotnet-install.ps1
+  powershell -NoProfile -ExecutionPolicy unrestricted -Command "%DotNet_Path%\dotnet-install.ps1 -InstallDir %DotNet_Path% -Version '%DotNet_Version%'"
+
+  if not exist "%DotNet%" (
+    call :print_error_message Could not install dotnet cli correctly. See '%Init_Tools_Log%' for more details.
+    exit /b 1
+  )
+
+  call %DotNet% --version
```
dpodder: double-quote `%DotNet%`

build.cmd
```diff
@@ -1,108 +1,220 @@
...
+
+:build_tests_simpleharness
+setlocal
+  cd %~dp0tests\simpleharness
+  call :dotnet_build  || exit /b 1
+  net.exe session 1>nul 2>&1 || (
+    call :print_error_message Cannot run simpleharness test because this is not an administrator window
+    exit /b 1
+  )
+  call %DotNet% run -c %BuildConfiguration% bin\%BuildConfiguration%\netcoreapp1.0\simpleharness.dll                                                                                                        || exit /b 1
```
dpodder: Suggest putting double quotes around parameters

build.cmd
```diff
@@ -1,108 +1,220 @@
...
+
+:dotnet_build
+  echo/
+  echo/  ==========
+  echo/   Building %cd%
+  echo/  ==========
+  call :remove_directory bin                                                    || exit /b 1
+  call :remove_directory obj                                                    || exit /b 1
+  call %DotNet% restore                                                         || exit /b 1
+  call %DotNet% build -c %BuildConfiguration% --version-suffix %VersionSuffix%  || exit /b 1
```
dpodder: Suggest double-quoting here too

build.cmd
```diff
@@ -1,108 +1,220 @@
...
+  echo/  ==========
+  echo/   Packing %cd%
+  echo/  ==========
+  call :remove_directory bin                                                                                                || exit /b 1
+  call :remove_directory obj                                                                                                || exit /b 1
+  call %DotNet% restore                                                                                                     || exit /b 1
+  call %DotNet% build -c %BuildConfiguration% --version-suffix %VersionSuffix%                                              || exit /b 1
+  call %DotNet% pack  -c %BuildConfiguration% --version-suffix %VersionSuffix% --output %OutputDirectory% --include-symbols || exit /b 1
+
+  :: FIXME: pack sources does not work with the current mixed version of the Tracing library (EXCEPTION THROWN).
```
dpodder: Has this been assigned a bug?

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -4,108 +4,95 @@
...
 using System;
 using System.Threading;
-using System.Reflection;
 using Xunit.Runners;
+using static Microsoft.Xunit.Performance.Api.XunitPerformanceLogger;
 
 namespace Microsoft.Xunit.Performance.Api
 {
-    public class XunitRunner
+    public static class XunitRunner
```
brianrob: Why does XunitRunner need to be static?  I can envision a world where multiple of these are created in a given process (sequentially).  It should be the case that nothing persists from the first one to the second one, and so doing this is likely to be easiest by having it be an instance.

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -4,108 +4,95 @@
...
-        private object _consoleLock = new object();
-
-        // Use an event to know when we're done
-        private ManualResetEvent _finished = new ManualResetEvent(false);
-
-        // Start out assuming success; we'll set this to 1 if we get a failed test
-        private int _result = 0;
-
-        public int Run(string assemblyPath)
+        public static int Run(string assemblyPath, string typeName = null)
```
brianrob: For clarity, can we convert the typeName change into an overload, perhaps named RunTestsForType(assemblyPath, typeName)?  Then you can also use the standard ArgumentOutOfRangeException(nameof(typeName)) instead of the exception text.

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -0,0 +1,300 @@
...
+                        (ulong)ClrTraceEventParser.Keywords.Loader |
+                        (ulong)ClrTraceEventParser.Keywords.Exception |
+                        (ulong)ClrTraceEventParser.Keywords.GC
+                    ),
+                }
+            };
+        }
+
+        /// <summary>
+        ///     1. In the specified assembly, get the EWT providers set as assembly attributes (PerformanceTestInfo)
```
dpodder: EWT -> ETW

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -4,108 +4,95 @@
...
         {
+            if(typeName != null && string.IsNullOrWhiteSpace(typeName))
+                throw new ArgumentException("The test type name cannot be white space.");
+
             // Create a runner for the specifid assembly.
-            using(AssemblyRunner runner = AssemblyRunner.WithoutAppDomain(assemblyPath))
+            using (var runner = AssemblyRunner.WithoutAppDomain(assemblyPath))
```
brianrob: From a code clarity POV, is there a reason that many of the static type names have been converted to var?  In the .NET Core repos, we avoid using var because it's not always obvious what the type name is.

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -4,108 +4,95 @@
...
+                    // Wait for tests to complete.
+                    finished.WaitOne();
                 }
 
-                // Return overall status.
-                return _result;
-            }
-        }
+                // Wait for the assembly runner to go idle.
+                using (var waitHandle = new ManualResetEvent(false))
```
brianrob: For clarity, can you please add { } around each of these?

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -4,108 +4,95 @@
...
+                    while (runner.Status != AssemblyRunnerStatus.Idle)
```
brianrob: Same here - { }.

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -4,108 +4,95 @@
...
-            lock (_consoleLock)
-            {
-                Console.WriteLine($"Running {info.TestCasesToRun} of {info.TestCasesDiscovered} tests...");
+                // Return overall status.
+                return result[0];
             }
         }
 
-        private void OnExecutionComplete(ExecutionCompleteInfo info)
+        private static void SetupRunnerCallbacks(AssemblyRunner runner, ManualResetEvent manualResetEvent, object consoleLock, int[] result)
```
brianrob: I believe that these arguments were originally instance variables, which would make this much cleaner.

src/xunit.performance.api/xunit.performance.api.csproj
```diff
@@ -1,36 +1,41 @@
 ﻿<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
+  <PropertyGroup Label="Configuration">
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+  </PropertyGroup>
   <PropertyGroup>
     <VersionPrefix>1.0.0-alpha</VersionPrefix>
-    <TargetFrameworks>netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>netstandard1.5</TargetFrameworks>
```
brianrob: Just want to make sure that the netstandard version change will still allow the runner to be used on desktop.

tests/simpleharness/Program.cs
```diff
@@ -1,68 +1,70 @@
...
     {
-        using (XunitPerformanceHarness p = new XunitPerformanceHarness(args))
+        public static void Main(string[] args)
         {
-            string entryAssemblyPath = Assembly.GetEntryAssembly().Location;
-            for(int i=0; i<5; i++)
-            {
-                p.RunBenchmarks(entryAssemblyPath);
-            }
+            using (var p = new XunitPerformanceHarness(args))
```
brianrob: Please add  { }.

tests/simpleharness/simpleharness.csproj
```diff
@@ -1,28 +1,26 @@
...
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.App">
-      <Version>1.0.1</Version>
-    </PackageReference>
-    <PackageReference Include="Microsoft.NET.Sdk">
-      <Version>1.0.0-alpha-20161104-2</Version>
+    <EmbeddedResource Include="app.manifest" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.App" Version="1.0.1" />
```
brianrob: Is this change required to use the version of CLI that is used by build.cmd?  Just want to make sure that this isn't an artifact of using a machine-wide CLI install.

src/xunit.performance.api/ConsoleOutput/XunitPerformanceLogger.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal static class XunitPerformanceLogger
```
brianrob: Can you add a class-level comment on what this is used for (and where it is used)?

src/xunit.performance.api/ConsoleDiagnosticMessageSink.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal sealed class ConsoleDiagnosticMessageSink : TestMessageVisitor<IDiagnosticMessage>
```
brianrob: Can you add a class-level comment on what this is used for?

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -0,0 +1,300 @@
...
+using System.Reflection;
+using System.Text;
+using Xunit;
+using Xunit.Abstractions;
+using static Microsoft.Xunit.Performance.Api.Native.Windows.Kernel32;
+using static Microsoft.Xunit.Performance.Api.XunitPerformanceLogger;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal static class ETWProfiler
```
brianrob: We already have a version of this code elsewhere in xunit-performance.  Is it possible to use it directly (or modify it slightly)?  It would be a shame to have two ETW implementations in the same project.

BuildForPublication.cmd
jorive: @brianrob This script is build for publication, and the idea is to have a clean repo when you want to publish a new package.

---
In reply to: [100207679](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100207679) [](ancestors = 100207679)

build.cmd
```diff
@@ -1,108 +1,220 @@
 @echo off
-goto :main
+@if defined _echo echo on
 
 :main
+setlocal enabledelayedexpansion
```
jorive: @brianrob I just added error checkins and fail fast (e.g. stop running subsequent steps if a project did not build, remove redundancy, and easier to reuse from other scripts such as buildapi)

---
In reply to: [100207724](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100207724) [](ancestors = 100207724)

netci.groovy
```diff
@@ -44,7 +44,6 @@ def projectFolder = Utilities.getFolderName(project) + '/' + Utilities.getFolder
                     }
 
                     shell("unpacker LocalPackages.pack .")
-                    shell("chmod +x BuildCliComponents.sh")
```
jorive: @brianrob No, as I changed the file attribute to executable in this PR.

---
In reply to: [100207823](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100207823) [](ancestors = 100207823)

src/xunit.performance.api/ETWProfiler.cs
jorive: @brianrob This is actually cleaner with less dependencies on the classes used by the runner and related projects. It also wraps the call to the xunit api.

---
In reply to: [100390285](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100390285) [](ancestors = 100390285)

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -0,0 +1,300 @@
...
+                sb.AppendLine($"  Interval :    {kvp.Value.Interval}");
+                sb.AppendLine($"  MaxInterval : {kvp.Value.MaxInterval}");
+                sb.AppendLine($"  MinInterval : {kvp.Value.MinInterval}");
+                sb.AppendLine();
+            }
+
+            WriteDebugLine(sb.ToString());
+        }
+
+        private sealed class PerformanceTestMessage
```
jorive: @brianrob This is just a port of the xunit runner, and it's used in the class below (Lines 229, 232, 253). I can probably do some cleanup here.

---
In reply to: [100344105](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100344105) [](ancestors = 100344105)

tests/simpleharness/app.manifest
```diff
@@ -0,0 +1,76 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jorive: This is the manifest that is usually used in .NET Framework to elevate your application to run as Administrator, but it does not work on .NET Core (at least for now). I figured I would give it a try. I can remove it if you think it is not necessary.

---
In reply to: [100349646](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100349646) [](ancestors = 100349646)

BuildForPublication.cmd
brianrob: Makes sense.  What about uncommitted changes to the repo?  I'm not sure where the line is to be honest.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -1,88 +1,69 @@
...
             _args = args;
+            _disposed = false;
 
             // Set the run id.
-            Configuration.RunId = DateTimeOffset.UtcNow.ToString("yyyy-MM-dd-HH-mm-ss");
+            Configuration.RunId = $"{DateTimeOffset.UtcNow.ToString("yyyyMMddHHmmss")}";
 
             // Set the file log path.
             // TODO: Conditionally set this based on whether we want a csv file written.
             Configuration.FileLogPath = Configuration.RunId + ".csv";
```
jorive: I think this should go away eventually.

---
In reply to: [100348024](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100348024) [](ancestors = 100348024)

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -1,88 +1,69 @@
...
-        private string[] _args;
-
         public XunitPerformanceHarness(string[] args)
         {
             _args = args;
+            _disposed = false;
 
             // Set the run id.
-            Configuration.RunId = DateTimeOffset.UtcNow.ToString("yyyy-MM-dd-HH-mm-ss");
+            Configuration.RunId = $"{DateTimeOffset.UtcNow.ToString("yyyyMMddHHmmss")}";
```
jorive: I can undo this change, but eventually this might be a parameter passed by the user like the runner. At least that is something @DrewScoggins mentioned, because it's easier to parse and do post processing on it.


---
In reply to: [100347473](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100347473) [](ancestors = 100347473)

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -1,88 +1,69 @@
...
-            // Invoke xunit to run benchmarks in the specified assembly.
-            XunitRunner runner = new XunitRunner();
-            runner.Run(assemblyPath);
         }
 
         private void ProcessResults()
         {
-            CSVMetricReader reader = new CSVMetricReader(Configuration.FileLogPath);
-            WriteStatisticsToFile(reader);
+            var reader = new CSVMetricReader(Configuration.FileLogPath);
```
jorive: The whole file output produced by the API needs some cleaning.
Currently, the ETW class does a checking on the OS, and only captures ETW on Windows.

---
In reply to: [100352115](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100352115) [](ancestors = 100352115)

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -1,88 +1,69 @@
...
         {
-            get { return BenchmarkConfiguration.Instance; }
+            Validate(assemblyPath);
+            int errCode = 0;
+            ETWProfiler.Profile(assemblyPath, Configuration.RunId, () => { errCode = XunitRunner.Run(assemblyPath); });
+            return errCode;
         }
 
-        public void RunBenchmarks(string assemblyPath)
+        private static void Validate(string assemblyPath)
```
jorive: Yes, it could be on the RunBenchmarks method. I just like to have the validation code in a separate method. I think it's cleaner and moving forward we will be getting more parameters that need validation (e.g. args[]).

---
In reply to: [100351798](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100351798) [](ancestors = 100351798)

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -94,27 +75,164 @@ private void WriteStatisticsToFile(CSVMetricReader reader)
...
+        {
+            // Close the log when all test cases have completed execution.
+            // HACK: This is a hack because we haven't found a way to close the file from within xunit.
+            BenchmarkEventSource.Log.Close();
+
+            // Process the results now that we know we're done executing tests.
+            ProcessResults();
+        }
+
+        private void FreeUnManagedResources()
```
jorive: No, I will remove it.

---
In reply to: [100352500](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100352500) [](ancestors = 100352500)

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -94,27 +75,164 @@ private void WriteStatisticsToFile(CSVMetricReader reader)
...
-                    r[col3_average] = avg.ToString();
-                    r[col4_stdevs] = stdev_s.ToString();
-                    r[col5_min] = min.ToString();
-                    r[col6_max] = max.ToString();
-                }
+                    yield return (testCaseName, measurement.Key, measurement.Value);
             }
+        }
+
+        private static DataTable CreateStatistics(CSVMetricReader reader)
```
jorive: Yes, I agree. The whole file output story of the API needs some cleaning, and this will not be here moving forward. This was just to unblock corefxlabs and provide their feature requests.

---
In reply to: [100353294](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100353294) [](ancestors = 100353294)

tests/simpleharness/simpleharness.csproj
jorive: This is a changed that the CLI version used by xunit.performance project understands.

---
In reply to: [100389849](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100389849) [](ancestors = 100389849)

src/xunit.performance.api/xunit.performance.api.csproj
jorive: From [https://docs.microsoft.com/en-us/dotnet/articles/standard/library](https://docs.microsoft.com/en-us/dotnet/articles/standard/library),
.netstandard1.5 support .NET Framework 4.6.2

---
In reply to: [100389083](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100389083) [](ancestors = 100389083)

src/xunit.performance.api/XunitRunner.cs
jorive: Yes, but the current request is to pass a string name of the benchmark to run. We can overload later.

---
In reply to: [100388074](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100388074) [](ancestors = 100388074)

build.cmd
jorive: Not that I know of.

---
In reply to: [100385566](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100385566) [](ancestors = 100385566)

src/xunit.performance.api/XunitRunner.cs
jorive: I've been using var throughout the code I've written as a shortcut to avoid typing (trying to be consistent). I could change it back. Thoughts?

---
In reply to: [100388288](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100388288) [](ancestors = 100388288)

src/xunit.performance.api/XunitRunner.cs
jorive: I agree with your vision. As runner is just a wrapper on top of xunit, I felt that it was enough being static plus you can still have a list of callbacks (in this case) running sequentially and potentially in parallel (e.g. tests running in different cores to maximize total throughput). These changes reduce the chances to have threading problems in the future because of shared states.

---
In reply to: [100387389](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100387389) [](ancestors = 100387389)

src/xunit.performance.api/CSVMetricReader.cs
```diff
@@ -12,12 +12,11 @@
...
 namespace Microsoft.Xunit.Performance.Api
 {
-    internal class CSVMetricReader
+    internal sealed class CSVMetricReader
     {
-        private Dictionary<string, List<Dictionary<string, double>>> _values = new Dictionary<string, List<Dictionary<string, double>>>();
-
         public CSVMetricReader(string csvPath)
         {
+            _values = new Dictionary<string, List<Dictionary<string, double>>>();
```
jorive: Preference and consistency with anything else inside this project. I can undo it :)

---
In reply to: [100209662](https://github.com/Microsoft/xunit-performance/pull/183#discussion_r100209662) [](ancestors = 100209662)

BuildForPublication.cmd
jorive: I could assume git is on the path and perform some git commands. (thoughts?)

jorive: I could assume git is on the path and perform some git commands. (thoughts?)

src/xunit.performance.api/Table/DataTable.cs
```diff
@@ -38,21 +37,24 @@ public IEnumerable<Row> Rows
             get { return _Rows; }
         }
 
-        public void WriteToCSV(string fullFilePath)
+        public void WriteToCSV(string fullFilePath, bool sort = true)
```
jorive: Agree. This was the fastest way to give corefxlab the feature.
I still want to refactor all the output story for the API, so I would like to do this in the near future.

netci.groovy
brianrob: Great.  Thanks.

src/xunit.performance.api/Table/DataTable.cs
brianrob: Ok, can you please file an issue for this?

src/xunit.performance.api/XunitPerformanceHarness.cs
brianrob: I think it's fine either way.  Just wanted to understand the reasoning.

src/xunit.performance.api/XunitPerformanceHarness.cs
brianrob: Does that mean that even with ETW, we're depending on the csv to provide execution time measurements?

src/xunit.performance.api/xunit.performance.api.csproj
brianrob: Perfect, thanks.

tests/simpleharness/simpleharness.csproj
brianrob: Great, thanks.

BuildForPublication.cmd
brianrob: I think that doing git clean -fxd is probably the right thing, but I also think that we should prompt the user with a warning before doing it.  It would be easy for someone to run BuildForPublication without knowing that it's going to blow away their untracked changes.

brianrob: Then you can just say that if your goal is to iterate on a change or do a private build, use build.cmd.

build.cmd
brianrob: Gotcha.  Ok, thanks.

BuildForPublication.cmd
tannergooding: I would think it more reasonable here to require the user to explicitly perform the cleaning themselves.

That is to say, if there are untracked changes or if there are artifacts where there shouldn't be, then fail the script with an appropriate error message and let the user resolve it themselves.

Otherwise, just let the script always run and give a warning under the above scenarios (which I think is the even better case -- perhaps the user wants to test some things before committing, which is a completely reasonable scenario).

jorive: I have added better handling of non-clean repos on the script.

src/xunit.performance.api/CSVMetricReader.cs
brianrob: Ok, no worries.

src/xunit.performance.api/ETWProfiler.cs
brianrob: Gotcha - so this is just part of the ETW configuration discovery.

src/xunit.performance.api/XunitRunner.cs
brianrob: Thinking about this a little bit further - what's the expectation on the API user?  I think that the user should not have to plumb the typename all the way through to this API.  It's not that it's a lot of work - it's just that it is more work that the user won't know they have to do.  What I'd like to see us do is capture the typename through the constructor of the harness, which take args, and then just filter calls to RunBenchmarks based on that (xunit can do this work).  That way, we're building features like this into the system that just light up rather than making changes that require changes to end-user code.  What do you think?

src/xunit.performance.api/XunitRunner.cs
brianrob: I think when you see the type right there (e.g. new <your type here>()), then it's not a big deal.  For other places where a method is called, (like line 27), we should have the type name present since it's not clear what the return type is.  What do you think?

tests/simpleharness/app.manifest
brianrob: Let's make sure it's required when running on desktop before we include it.

src/xunit.performance.api/ETWProfiler.cs
brianrob: Per offline conversation, let's create an issue to track merging the two implementations.

src/xunit.performance.api/XunitRunner.cs
brianrob: Let's file an issue for this.

tests/simpleharness/app.manifest
brianrob: Per offline conversation, we'll remove this for now.  Let's file an issue to make sure that desktop processes get elevated properly.

src/xunit.performance.api/ETWProfiler.cs
jorive: https://github.com/Microsoft/xunit-performance/issues/187

src/xunit.performance.api/Table/DataTable.cs
jorive: https://github.com/Microsoft/xunit-performance/issues/188

src/xunit.performance.api/XunitPerformanceHarness.cs
jorive: In this PR, we only generate etl files and markdown summary, but we do not read the etl to get more information.

tests/simpleharness/app.manifest
jorive: https://github.com/Microsoft/xunit-performance/issues/189

src/xunit.performance.api/XunitRunner.cs
jorive: https://github.com/Microsoft/xunit-performance/issues/190

src/xunit.performance.api/XunitRunner.cs
jorive: This is not used in this PR and API users have not access to it directly, as they can only call the harness.

src/xunit.performance.api/XunitRunner.cs
jorive: Makes sense.

src/xunit.performance.api/XunitRunner.cs
jorive: https://github.com/Microsoft/xunit-performance/issues/190

BuildForPublication.cmd
```diff
@@ -21,7 +21,7 @@ setlocal
     choice /T 10 /D N /C YN /M "WARNING: Repository is not clean. Press Y to continue or N to stop. "
```
brianrob: May be worthwhile to change the text slightly to be a bit more clear - something like "WARNING: The repo contains uncommitted changes and you're building for publication."  Something along those lines.

brianrob:
@jorive, thanks for making changes and filing issues.  I have one additional question about BuildForPublication.cmd, but otherwise, LGTM.  Can you please squash this down to the number of commits that you think makes sense based on the work items contained in the PR?

PR #191, opened 2017-02-15
Adding BenchView data generation (initial PR)

jorive:
Read generated *.etl file and generates xml file format understood by BenchView
(Currently disabled file generation on Linux because the csv file is opened until harness shuts down and this does not allow reading before hands. I'm currently working on a fix).


.gitignore
```diff
@@ -223,3 +223,5 @@ nuget.exe
 *.nuget.props
 *.nuget.targets
 *.etl
+*.xml
+*.md
```
dpodder: Are we sure we can add `*.xml` and `*.md` to .gitignore at the repo root? Wouldn't this be too broad?

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -61,29 +74,28 @@ static ETWProfiler()
...
         {
+            const int bufferSizeMB = 128;
             var sessionFileName = $"{sessionName}-{Path.GetFileNameWithoutExtension(assemblyFileName)}";
-            var userFileName = $"{sessionFileName}.etl";
-            var kernelFileName = $"{sessionFileName}.kernel.etl";
-            var currentWorkingDirectory = Directory.GetCurrentDirectory();
-            var userFullFileName = Path.Combine(currentWorkingDirectory, userFileName);
-            var kernelFullFileName = Path.Combine(currentWorkingDirectory, kernelFileName);
+            var userFullFileName = Path.Combine(outputDirectory, $"{sessionFileName}.etl");
+            var kernelFullFileName = Path.Combine(outputDirectory, $"{sessionFileName}.kernel.etl"); /* without this parameter, EnableKernelProvider will fail */
```
dpodder: \<nit\>Can we use `//comment` instead of `/*comment*/` here?\</nit\>

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -107,9 +119,196 @@ public static void Profile(string assemblyFileName, string sessionName, Action a
...
+        }
+
+        private static EtwPerformanceMetricEvaluationContext GetEtwReader(
+            string fileName,
+            string sessionName,
+            IEnumerable<PerformanceTestMessage> performanceTestMessages)
+        {
+            using (var source = new ETWTraceEventSource(fileName))
+            {
+                if (source.EventsLost > 0)
```
dpodder: When do source events get counted as "lost"?

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -107,9 +119,196 @@ public static void Profile(string assemblyFileName, string sessionName, Action a
...
+                    xmlIterationsElement.Add(xmlIterationElement);
+                }
+            }
+
+            var xmlDoc = new XDocument(xmlAssembliesElement);
+            using (var xmlFile = File.Create(xmlFileName))
+            {
+                xmlDoc.Save(xmlFile, SaveOptions.DisableFormatting);
+            }
+            WriteInfoLine($"XML BenchView tests saved to \"{xmlFileName}\"");
```
dpodder: I wouldn't call these "BenchView" tests--BenchView doesn't understand XML. 😄 Could we change this to say something like, `$"Performance results saved to \"{xmlFileName}\""`?

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -107,9 +119,196 @@ public static void Profile(string assemblyFileName, string sessionName, Action a
...
+                        "iteration",
+                        xmlAttributes.ToArray());
+                    xmlIterationsElement.Add(xmlIterationElement);
+                }
+            }
+
+            var xmlDoc = new XDocument(xmlAssembliesElement);
+            using (var xmlFile = File.Create(xmlFileName))
+            {
+                xmlDoc.Save(xmlFile, SaveOptions.DisableFormatting);
```
dpodder: How much space do we save by disabling formatting? I'm curious whether the tradeoff is worth it (since we lose "readability" when debugging issues). Sadly the files can get long enough to defeat most auto-formatters...

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -156,7 +361,13 @@ private static void SetPreciseMachineCounters(IEnumerable<ProviderInfo> provider
                 }
 
                 if (profileSourceIDs.Count > 0)
+                {
+                    /*
+                     * FIXME: This function changes the -pmcsources intervals machine wide.
+                     *  Maybe we should undo/revert these changes!
+                     */
```
dpodder: Could we file an issue for the FIXME?

dpodder: Also, \<nit\>We should use `//comments` over `/*comments*/`\</nit\>.

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -142,6 +341,12 @@ private static void SetPreciseMachineCounters(IEnumerable<ProviderInfo> provider
         {
             if (IsWindows8OrGreater)
             {
+                /*
+                 *  TODO: Add default BranchMispredictions,CacheMisses flags (if available)
+                 *    1. This reads as: flags read from config file in the case of Windows!
+                 *    2. Format { "Name" : "BranchMispredictions", "interval" : "100000" }
+                 */
```
dpodder: \<nit\>Please use `//comments` instead of `/*comments*/`\</nit\>.

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -170,10 +381,8 @@ private static bool NeedSeparateKernelSession(KernelProviderInfo kernelProviderI
                 return true;
 
             // CPU counters need the special kernel session
-            if (((KernelTraceEventParser.Keywords)kernelProviderInfo.Keywords & KernelTraceEventParser.Keywords.PMCProfile) != 0)
-                return true;
-
-            return false;
+            var keywords = (KernelTraceEventParser.Keywords)kernelProviderInfo.Keywords & KernelTraceEventParser.Keywords.PMCProfile;
+            return (keywords != 0) ? true : false;
```
dpodder: `return (keywords != 0);`

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -107,9 +119,196 @@ public static void Profile(string assemblyFileName, string sessionName, Action a
...
+            var mdTable = MarkdownHelper.GenerateMarkdownTable(dt);
+            MarkdownHelper.Write(mdFileName, mdTable);
+
+            collectOutputFilesCallback(mdFileName);
+            Console.WriteLine(mdTable);
+        }
+
+        private static DataTable CreateStatistics(AssemblyModel assemblyModel)
+        {
+            var dt = new DataTable();
```
dpodder: What happens if we add an extra column, or forget to add a column, compared to what MarkdownHelper is expecting? Is there any way to typecheck the columns we add and their respective data types?

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal sealed class AssemblyModel
```
dpodder: Could this be auto-serialized, instead of manually generating the XmlNodes and adding children to them like we currently do?

src/xunit.performance.api/ETWProfiler.cs
jorive: When the ETW 'recording' exceeds the buffer size. This piece of the code was ported from old runner, but I would prefer to revisit it in the near future.

---
In reply to: [101316411](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r101316411) [](ancestors = 101316411)

src/xunit.performance.api/ETWProfiler.cs
jorive: [https://github.com/Microsoft/xunit-performance/issues/192](https://github.com/Microsoft/xunit-performance/issues/192)

---
In reply to: [101318647](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r101318647) [](ancestors = 101318647)

src/xunit.performance.api/ETWProfiler.cs
jorive: Currently this is hardcoded, but I'm planning on fixing this in future PR.

---
In reply to: [101323783](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r101323783) [](ancestors = 101323783)

src/xunit.performance.api/Model/AssemblyModel.cs
jorive: I will look into this in future PR.

---
In reply to: [101324384](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r101324384) [](ancestors = 101324384)

src/xunit.performance.api/ETWProfiler.cs
dpodder: Gotcha, thanks. Per our offline conversation, I think this is good to keep as-is for now.

---
In reply to: [101368446](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r101368446) [](ancestors = 101368446,101316411)

src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                if (_currentProcesses.Contains(args.ParentID))
+                    _currentProcesses.Add(args.ProcessID);
+            };
+            benchmarkParser.Source.Kernel.ProcessStop += delegate (ProcessTraceData args)
+            {
+                _currentProcesses.Remove(args.ProcessID);
+            };
+
+            //benchmarkParser.Source.Kernel.All += traceEvent =>
```
brianrob: Probably can just get rid of this?

src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.Xunit.Performance.Sdk;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using static Microsoft.Xunit.Performance.Api.XunitPerformanceLogger;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal class EtwPerformanceMetricEvaluationContext : PerformanceMetricEvaluationContext, IDisposable, IPerformanceMetricReader
```
brianrob: Is this file a straight copy/paste or does it have changes?

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -142,6 +284,12 @@ private static void SetPreciseMachineCounters(IEnumerable<ProviderInfo> provider
         {
             if (IsWindows8OrGreater)
             {
+                //
+                // TODO: Add default BranchMispredictions,CacheMisses flags (if available)
```
brianrob: I think you already added this above?

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -156,7 +304,13 @@ private static void SetPreciseMachineCounters(IEnumerable<ProviderInfo> provider
                 }
 
                 if (profileSourceIDs.Count > 0)
+                {
+                    // 
+                    // FIXME: This function changes the -pmcsources intervals machine wide.
```
brianrob: Yes, we should fix this in this PR.  I don't think we want machine-wide state changes not getting cleaned up.

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -107,9 +116,142 @@ public static void Profile(string assemblyFileName, string sessionName, Action a
 
             TraceEventSession.MergeInPlace(userFullFileName, Console.Out);
             WriteInfoLine($"ETW Tracing Session saved to \"{userFullFileName}\"");
+            collectOutputFilesCallback(userFullFileName);
+
+            var assemblyModel = GetAssemblyModel(assemblyFileName, userFullFileName, sessionName, performanceTestMessages);
```
brianrob: Is the AssemblyModel an Xunit specific construct?  I don't think we want anything xunit specific here, given that we're going to have tests that don't execute under xunit.

jorive: This is not really a Xunit specific construct (more like a copied schema/model of its output). This is a type used to serialized collected data into something that BenchView tools already parse. In addition, we would not break backward compatibility by switching to the new API (without creating a new format on the BenchView tools to parse). I am opened to redefine this later on if needed.

src/xunit.performance.api/ETWProfiler.cs
jorive: I do not think this is a new issue with the API. Can we change this in subsequent PR?

src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
jorive: Almost a straight copy from the *run project (net46). I believe we should break down projects and do general cleanup, but that should be outside this PR. At the same time, it might be good to have a class we can extend without worrying about breaking "legacy code"?

src/xunit.performance.api/ETWProfiler.cs
brianrob: Ok, no worries.  Just want to make sure that we didn't put any xunit specifics in here.

---
In reply to: [102335366](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r102335366) [](ancestors = 102335366)

brianrob:
>                     action.Invoke();

What happens if something throws in here?  The session will certainly get cleaned up, but what will happen to the process?  Just want to make sure that automated processes will not completely choke.  Do we need to wrap this in a try/catch?

---
Refers to: src/xunit.performance.api/ETWProfiler.cs:113 in 9264a48. [](commit_id = 9264a48c6fe45b07a497f70a53880e74a7f0883d, deletion_comment = False)

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -107,9 +116,142 @@ public static void Profile(string assemblyFileName, string sessionName, Action a
...
+            collectOutputFilesCallback(mdFileName);
+            Console.WriteLine(mdTable);
+
+            var csvFileName = Path.Combine(outputDirectory, $"{fileNameWithoutExtension}.csv");
+            dt.WriteToCSV(csvFileName);
+            WriteInfoLine($"Statistics written to \"{csvFileName}\"");
+            collectOutputFilesCallback(csvFileName);
+        }
+
+        private static DataTable CreateStatistics(AssemblyModel assemblyModel)
```
brianrob: I take it this is covered under the outstanding issue to clean-up the output generation code?  This looks like it belongs in a different file and not as part of the ETW profiling code.

src/xunit.performance.api/ETWProfiler.cs
brianrob: Ok, if not a new issue, can you please file an issue for this?

---
In reply to: [102335560](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r102335560) [](ancestors = 102335560)

src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
brianrob: Let's see if we can deprecate the older stuff soon, and maybe have a branch that has the deprecated stuff in it so that if we have to hotfix it we can.  Then, we can delete the old stuff out of master, and not even have to worry about it.

---
In reply to: [102346911](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r102346911) [](ancestors = 102346911)

jorive: That sounds great!

---
In reply to: [102349184](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r102349184) [](ancestors = 102349184,102346911)

src/xunit.performance.api/ETWProfiler.cs
jorive: I am moving it to a different place on the next PR. It cleans up some of the File IO generated by the API and duplicated code

---
In reply to: [102348467](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r102348467) [](ancestors = 102348467)

jorive:
>                     action.Invoke();

Sessions will clean up, and the client will need to catch the exception. I'm not sure what would be best to do here. If we catch the exception, and hide it from the user, how do we let them know? We should talk about this one some more.

---
In reply to: [281518575](https://github.com/Microsoft/xunit-performance/pull/191#issuecomment-281518575) [](ancestors = 281518575)

---
Refers to: src/xunit.performance.api/ETWProfiler.cs:113 in 9264a48. [](commit_id = 9264a48c6fe45b07a497f70a53880e74a7f0883d, deletion_comment = False)

src/xunit.performance.api/ETWProfiler.cs
jorive: [https://github.com/Microsoft/xunit-performance/issues/192](https://github.com/Microsoft/xunit-performance/issues/192)

---
In reply to: [102348741](https://github.com/Microsoft/xunit-performance/pull/191#discussion_r102348741) [](ancestors = 102348741,102335560)

src/xunit.performance.api/GenericProfiler.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal static class GenericProfiler
```
brianrob: Do we even need this?  It looks like in the API could just call action.Invoke() directly.  There's no interface that we have to match.

brianrob:
>                     action.Invoke();

Ok, no worries for this PR.  Can you file an issue so we remember to look at this?  Don't block on this PR.

---
In reply to: [281523106](https://github.com/Microsoft/xunit-performance/pull/191#issuecomment-281523106) [](ancestors = 281523106,281518575)

---
Refers to: src/xunit.performance.api/ETWProfiler.cs:113 in 9264a48. [](commit_id = 9264a48c6fe45b07a497f70a53880e74a7f0883d, deletion_comment = False)

src/xunit.performance.api/GenericProfiler.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        public static void Record(string assemblyFileName, string sessionName, string outputDirectory, Action action, Action<string> collectOutputFilesCallback)
+        {
+            action.Invoke();
```
brianrob: Can you just call action.Invoke() directly from the Harness API?  You control the body of the delegate from which this is called, so nothing has to match the "interface".

brianrob:
:shipit:

PR #195, opened 2017-02-23
Adding a Ctrl-C, Ctrl-Break, Close handle for disposable objects.

jorive:
# The Problem:
When the running harness was stopped with a console control signal such as: Ctrl-C, Ctrl-Break, or Close signals, the trace event session did not clean up properly, the handles to the etl files and sessions themselves persisted.

# The Solution:
Wraps the trace event sessions in an object that subscribe them to a Ctrl listener, and dispose them accordingly when the signal is received.

## Note:
Killing the process under the debugger, with the signals specified above, does not make it very happy. The process seems continue running or crashes.

### DIFF
--- a/src/xunit.performance.api/ETWProfiler.cs
+++ b/src/xunit.performance.api/ETWProfiler.cs
@@ -15,6 +15,11 @@ namespace Microsoft.Xunit.Performance.Api
 {
     internal static class ETWProfiler
     {
+        private static SafeTerminateHandler<TraceEventSession> MakeSafeTerminateTraceEventSession(string sessionName, string fileName)
+        {
+            return new SafeTerminateHandler<TraceEventSession>(() => new TraceEventSession(sessionName, fileName));
+        }
+
         /// <summary>
         ///     1. In the specified assembly, get the ETW providers set as assembly attributes (PerformanceTestInfo)
         ///     2. Check if the benchmark assembly request Precise Machine Counters(PMC) to be collected
@@ -25,17 +30,18 @@ namespace Microsoft.Xunit.Performance.Api
         ///     7. Merge ETL files.
         /// </summary>
         /// <param name="assemblyFileName"></param>
-        /// <param name="sessionName"></param>
+        /// <param name="runId"></param>
         /// <param name="outputDirectory"></param>
         /// <param name="action"></param>
         /// <param name="collectOutputFilesCallback">Callback used to collect a list of files generated.</param>
         /// <returns></returns>
-        public static void Record(string assemblyFileName, string sessionName, string outputDirectory, Action action, Action<string> collectOutputFilesCallback)
+        public static void Record(string assemblyFileName, string runId, string outputDirectory, Action action, Action<string> collectOutputFilesCallback)
         {
             const int bufferSizeMB = 128;
-            var sessionFileName = $"{sessionName}-{Path.GetFileNameWithoutExtension(assemblyFileName)}";
-            var userFullFileName = Path.Combine(outputDirectory, $"{sessionFileName}.etl");
-            var kernelFullFileName = Path.Combine(outputDirectory, $"{sessionFileName}.kernel.etl"); // without this parameter, EnableKernelProvider will fail
+            var sessionName = $"Performance-Api-Session-{runId}";
+            var name = $"{runId}-{Path.GetFileNameWithoutExtension(assemblyFileName)}";
+            var userFullFileName = Path.Combine(outputDirectory, $"{name}.etl");
+            var kernelFullFileName = Path.Combine(outputDirectory, $"{name}.kernel.etl"); // without this parameter, EnableKernelProvider will fail
 
             PrintProfilingInformation(assemblyFileName, sessionName, userFullFileName);
 
@@ -43,8 +49,9 @@ namespace Microsoft.Xunit.Performance.Api
             var kernelProviderInfo = providers.OfType<KernelProviderInfo>().FirstOrDefault();
 
             var needKernelSession = NeedSeparateKernelSession(kernelProviderInfo);
-            using (var kernelSession = needKernelSession ? new TraceEventSession(KernelTraceEventParser.KernelSessionName, kernelFullFileName) : null)
+            using (var safeKernelSession = needKernelSession ? MakeSafeTerminateTraceEventSession(KernelTraceEventParser.KernelSessionName, kernelFullFileName) : null)
             {
+                var kernelSession = safeKernelSession.BaseDisposableObject;
                 if (kernelSession != null)
                 {
                     SetPreciseMachineCounters(providers);
@@ -54,8 +61,9 @@ namespace Microsoft.Xunit.Performance.Api
                     kernelSession.EnableKernelProvider(flags, stackCapture);
                 }
 
-                using (var userEventSession = new TraceEventSession(sessionName, userFullFileName))
+                using (var safeUserEventSession = MakeSafeTerminateTraceEventSession(sessionName, userFullFileName))
                 {
+                    var userEventSession = safeUserEventSession.BaseDisposableObject;
                     userEventSession.BufferSizeMB = bufferSizeMB;
 
                     var flags = KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.ImageLoad | KernelTraceEventParser.Keywords.Thread;
@@ -73,14 +81,13 @@ namespace Microsoft.Xunit.Performance.Api
             WriteInfoLine($"ETW Tracing Session saved to \"{userFullFileName}\"");
             collectOutputFilesCallback(userFullFileName);
 
-            var assemblyModel = GetAssemblyModel(assemblyFileName, userFullFileName, sessionName, performanceTestMessages);
-            var fileNameWithoutExtension = Path.GetFileNameWithoutExtension(userFullFileName);
-            var xmlFileName = Path.Combine(outputDirectory, $"{fileNameWithoutExtension}.xml");
+            var assemblyModel = GetAssemblyModel(assemblyFileName, userFullFileName, runId, performanceTestMessages);
+            var xmlFileName = Path.Combine(outputDirectory, $"{name}.xml");
             new AssemblyModelCollection { assemblyModel }.Serialize(xmlFileName);
             WriteInfoLine($"Performance results saved to \"{xmlFileName}\"");
             collectOutputFilesCallback(xmlFileName);
 
-            var mdFileName = Path.Combine(outputDirectory, $"{fileNameWithoutExtension}.md");
+            var mdFileName = Path.Combine(outputDirectory, $"{name}.md");
             var dt = assemblyModel.GetStatistics();
             var mdTable = MarkdownHelper.GenerateMarkdownTable(dt);
             MarkdownHelper.Write(mdFileName, mdTable);
@@ -88,7 +95,7 @@ namespace Microsoft.Xunit.Performance.Api
             collectOutputFilesCallback(mdFileName);
             Console.WriteLine(mdTable);
 
-            var csvFileName = Path.Combine(outputDirectory, $"{fileNameWithoutExtension}.csv");
+            var csvFileName = Path.Combine(outputDirectory, $"{name}.csv");
             dt.WriteToCSV(csvFileName);
             WriteInfoLine($"Statistics written to \"{csvFileName}\"");
             collectOutputFilesCallback(csvFileName);
--- a/src/xunit.performance.api/ETWProfiler.cs
+++ b/src/xunit.performance.api/ETWProfiler.cs
@@ -1,7 +1,6 @@
 ﻿using Microsoft.Diagnostics.Tracing;
 using Microsoft.Diagnostics.Tracing.Parsers;
 using Microsoft.Diagnostics.Tracing.Session;
-using Microsoft.Xunit.Performance.Api.Table;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot retest Windows / Ubuntu Debug Build and Windows / Ubuntu Release Build

jorive:
@dotnet-bot retest Windows / Ubuntu Release Build

src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -73,22 +81,21 @@ public static void Record(string assemblyFileName, string sessionName, string ou
...
-            var fileNameWithoutExtension = Path.GetFileNameWithoutExtension(userFullFileName);
-            var xmlFileName = Path.Combine(outputDirectory, $"{fileNameWithoutExtension}.xml");
+            var assemblyModel = GetAssemblyModel(assemblyFileName, userFullFileName, runId, performanceTestMessages);
+            var xmlFileName = Path.Combine(outputDirectory, $"{name}.xml");
             new AssemblyModelCollection { assemblyModel }.Serialize(xmlFileName);
             WriteInfoLine($"Performance results saved to \"{xmlFileName}\"");
             collectOutputFilesCallback(xmlFileName);
 
-            var mdFileName = Path.Combine(outputDirectory, $"{fileNameWithoutExtension}.md");
+            var mdFileName = Path.Combine(outputDirectory, $"{name}.md");
```
brianrob: It looks like you have some other file name changes in this PR too.  Are these intentional?

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Microsoft.Xunit.Performance.Api.Native.Windows;
+using System;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal sealed class SafeTerminateHandler<T> : IDisposable
```
brianrob: I think this class deserves a comment block and some line-level comments to describe what it is for and what it does.  It certainly is a helpful wrapper class, but it also encapsulates some rather complex code, which we should document.

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+                        BaseDisposableObject?.Dispose();
+                        break;
+                    case Kernel32.CtrlTypes.CTRL_LOGOFF_EVENT:
+                    case Kernel32.CtrlTypes.CTRL_SHUTDOWN_EVENT:
+                        break;
+                }
+                return false;
+            };
+
+            if (!Kernel32.SetConsoleCtrlHandler(_HandlerRoutine, true))
```
brianrob: Nit: Please add { }.

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Microsoft.Xunit.Performance.Api.Native.Windows;
+using System;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal sealed class SafeTerminateHandler<T> : IDisposable
+        where T : class, IDisposable
+    {
+        public SafeTerminateHandler(Func<T> createCallback)
```
brianrob: Can you please document the reason that you have a create callback vs. taking in the object itself?  I'm assuming that your goal is to make it more obvious that the lifetime of the object is owned by the SafeTerminateHandler instead of other code that might create the object and then dispose of it on its own.

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+        #region IDisposable Support
+        private bool _disposedValue = false; // To detect redundant calls
+
+        void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    // TODO: dispose managed state (managed objects).
```
brianrob: Is this TODO still needed?  You should only have to dispose of the base disposable object, right?

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+                    BaseDisposableObject?.Dispose();
+                }
+
+                // TODO: free unmanaged resources (unmanaged objects) and override a finalizer below.
```
brianrob: Do you need these TODOs?

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+            Dispose(false);
+        }
+
+        // This code added to correctly implement the disposable pattern.
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+            Dispose(true);
+            // TODO: uncomment the following line if the finalizer is overridden above.
```
brianrob: I don't think you need this TODO.

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+            BaseDisposableObject = null;
+
+            _HandlerRoutine = (Kernel32.CtrlTypes dwCtrlType) =>
+            {
+                switch (dwCtrlType)
+                {
+                    case Kernel32.CtrlTypes.CTRL_C_EVENT:
+                    case Kernel32.CtrlTypes.CTRL_BREAK_EVENT:
+                    case Kernel32.CtrlTypes.CTRL_CLOSE_EVENT:
+                        BaseDisposableObject?.Dispose();
```
brianrob: It looks to me like there is a race condition here where BaseDisposableObject?.Dispose() can be called multiple times - once from the handler, and then from an explicit call to Dispose and/or the finalizer.  I think you may need to plumb everything through Dispose, and perhaps even do an InterlockedCompareExchange on _disposedValue to make sure there aren't duplicate disposal operations.

src/xunit.performance.api/SafeTerminateHandler.cs
brianrob: I'd recommend calling this something like "SafeDisposableWrapper<T>" or something like that so that it's clear that it's a wrapper around a disposable object, and has semantics similar to "safe" objects like SafeHandle.

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+                // TODO: free unmanaged resources (unmanaged objects) and override a finalizer below.
+                // TODO: set large fields to null.
+                if (!Kernel32.SetConsoleCtrlHandler(_HandlerRoutine, false))
+                    throw new ObjectDisposedException("SetConsoleCtrlHandler failed to remove handler.");
+
+                _disposedValue = true;
+            }
+        }
+
+        // TODO: override a finalizer only if Dispose(bool disposing) above has code to free unmanaged resources.
```
brianrob: Do you need this TODO?

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: Yes, I will remove those TODOs. They are created by the refactoring tool of VS and I forgot to remove them.

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: Yes, I will document the class.

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: I wanted to make clear through the name that it's meant to handle terminate signals (currently only on Windows). Therefore, I think `SafeDisposableWrapper` would omit the fact that it disposes when a terminate signal is received by the application.
Thoughts?

src/xunit.performance.api/ETWProfiler.cs
jorive: Yes, this is intentional. I split the runid and session names. The session names have a prefix `Performance-Api`, so it is easier to find any session opened by the API and possibly left opened. The second one, the runid is used in conjunction with the assembly being measured, to create the output files of the API.

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: What is a guideline you use to add { }? Any block in C#?

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: You are right. I should "lock" this section. Any suggestion?

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+                switch (dwCtrlType)
+                {
+                    case Kernel32.CtrlTypes.CTRL_C_EVENT:
+                    case Kernel32.CtrlTypes.CTRL_BREAK_EVENT:
+                    case Kernel32.CtrlTypes.CTRL_CLOSE_EVENT:
+                        BaseDisposableObject?.Dispose();
+                        break;
+                    case Kernel32.CtrlTypes.CTRL_LOGOFF_EVENT:
+                    case Kernel32.CtrlTypes.CTRL_SHUTDOWN_EVENT:
+                        break;
```
dpodder: Is it required to add these if they're not treated specifically? Can we use `default` instead?

src/xunit.performance.api/SafeTerminateHandler.cs
dpodder: What happens on Linux? Does this need to go into an `#if WINDOWS` block of some sort?

src/xunit.performance.api/SafeTerminateHandler.cs
dpodder: Wow, good catch @brianrob!

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            if (!Kernel32.SetConsoleCtrlHandler(_HandlerRoutine, true))
+                throw new Exception("SetConsoleCtrlHandler failed to add handler.");
+
+            BaseDisposableObject = createCallback();
+        }
+
+        public T BaseDisposableObject { get; }
+
+        private readonly Kernel32.PHANDLER_ROUTINE _HandlerRoutine;
```
dpodder: This is for Windows only, right? Can we call it `_WindowsHandlerRoutine` to differentiate it vs. a future Linux or OSX implementation--or at least indicate that it's intended for Windows only, even if Linux/OSX don't need similar handlers?

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: Currently the focus of the API has been around Windows, but I will start working on the Linux part soon. In addition, this class is used inside the ETW class which only runs on Windows platform.
I would like to delay it to a future PR.

src/xunit.performance.api/SafeTerminateHandler.cs
jorive: Same as above.
I agree, but I would rather make the split (Windows/Linux) somewhere using a façade to keep the code platform agnostic.

src/xunit.performance.api/SafeTerminateHandler.cs
dpodder: OK, I mistakenly thought we already supported Linux. I think I got this confused with another repo--my bad. 😕 Sounds good, this can be done later.

brianrob: I was going to say yes any block in C#, but I stand corrected by the coding guidelines: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

src/xunit.performance.api/SafeTerminateHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+                    case Kernel32.CtrlTypes.CTRL_CLOSE_EVENT:
+                        Dispose();
+                        break;
+                    default:
+                        break;
+                }
+                return false;
+            };
+
+            if (!Kernel32.SetConsoleCtrlHandler(_HandlerRoutine, true))
```
dpodder: Is this safe outside of the critical section? It's possible for the console ctrl handler to get created, but ctrl-C to be invoked before we enter the critical section below to set _disposedValue to false, and we never remove the handler. Even if this is a "safe" leak (the handler goes away when the process exits) I think it would better form to move this into the critical section.

src/xunit.performance.api/Kernel32.cs
```diff
@@ -87,5 +87,22 @@ private enum ConditionMask : byte
...
+            CTRL_C_EVENT = 0,
+            CTRL_BREAK_EVENT = 1,
+            CTRL_CLOSE_EVENT = 2,
+            CTRL_LOGOFF_EVENT = 5,
+            CTRL_SHUTDOWN_EVENT = 6
+        }
+
+        public delegate bool PHANDLER_ROUTINE(CtrlTypes dwCtrlType);
+
+        [DllImport("kernel32.dll", SetLastError = true, CallingConvention = CallingConvention.Winapi)]
```
stephentoub: Why are we using P/Invokes rather than the CoreFx support that wraps this, e.g. Console.CancelKeyPress?  Is there something we're missing in the Console implementation (which works cross-platform) that this needs?

jorive: @stephentoub From what I understand, the Console.CancelKeyPress does not handle closing the console window. I could not find documentation about it, and I tried debugging a example and I did not seem to get the event. Could you please verify?

PR #196, opened 2017-02-28
New metrics parsing + bug fixes

jorive:
1. Doubled the trace event session's object buffer size
2. Parse Pmc default collected data (Branch misspredictions + Cache misses)
3. Fix a bug on the xunit runner that was not filtering on Benchmark and was running all Xunit Facts.

### DIFF
--- a/src/xunit.performance.api/XunitBenchmark.cs
+++ b/src/xunit.performance.api/XunitBenchmark.cs
@@ -14,22 +14,24 @@ namespace Microsoft.Xunit.Performance.Api
             // TODO: Read this from file.
             //   1. This reads as: flags read from config file based on running platform!
             //   2. Format could be something like this: `{ "Name" : "BranchMispredictions", "interval" : "100000" }`
-            RequiredProviders = new ProviderInfo[]
+            PerformanceMonitorCounters = new List<IPerformanceMonitorCounter>
+            {
+                new BranchMispredictionsPerformanceMonitorCounter(),
+                new CacheMissesPerformanceMonitorCounter(),
+            };
+
+            PerformanceMetricInfos = new List<BasePerformanceMonitorCounter>
+            {
+                new GenericPerformanceMonitorCounterMetric<BranchMispredictionsPerformanceMonitorCounter>(new BranchMispredictionsPerformanceMonitorCounter()),
+                new GenericPerformanceMonitorCounterMetric<CacheMissesPerformanceMonitorCounter>(new CacheMissesPerformanceMonitorCounter()),
+            };
+
+            RequiredProviders = new List<ProviderInfo>
             {
                 new KernelProviderInfo()
                 {
                     Keywords = (ulong)KernelTraceEventParser.Keywords.Process | (ulong)KernelTraceEventParser.Keywords.Profile,
                     StackKeywords = (ulong)KernelTraceEventParser.Keywords.Profile
-                },
-                new CpuCounterInfo()
-                {
-                    CounterName = "BranchMispredictions",
-                    Interval = 100000
-                },
-                new CpuCounterInfo()
-                {
-                    CounterName = "CacheMisses",
-                    Interval = 100000
                 },
                 new UserProviderInfo()
                 {
@@ -52,6 +54,15 @@ namespace Microsoft.Xunit.Performance.Api
                     ),
                 }
             };
+
+            foreach (var pmc in PerformanceMonitorCounters)
+            {
+                RequiredProviders.Add(new CpuCounterInfo()
+                {
+                    CounterName = pmc.Name,
+                    Interval = pmc.Interval,
+                });
+            }
         }
 
         public static (IEnumerable<ProviderInfo> providers, IEnumerable<PerformanceTestMessage> performanceTestMessages) GetMetadata(string assemblyFileName)
@@ -78,11 +89,26 @@ namespace Microsoft.Xunit.Performance.Api
                         select provider;
 
                     providers = ProviderInfo.Merge(providers);
+
+                    // Inject implicit pmc counters
+                    foreach (var test in testMessageVisitor.Tests)
+                    {
+                        var metrics = new List<PerformanceMetricInfo>(test.Metrics);
+                        foreach (var performanceMetricInfo in PerformanceMetricInfos)
+                            if (performanceMetricInfo.IsValidPmc)
+                                metrics.Add(performanceMetricInfo);
+                        test.Metrics = metrics;
+                    }
+
                     return (ProviderInfo.Merge(RequiredProviders.Concat(providers)), testMessageVisitor.Tests);
                 }
             }
         }
 
-        public static IEnumerable<ProviderInfo> RequiredProviders { get; }
+        public static List<ProviderInfo> RequiredProviders { get; }
+
+        private static List<IPerformanceMonitorCounter> PerformanceMonitorCounters { get; }
+
+        private static List<BasePerformanceMonitorCounter> PerformanceMetricInfos { get; }
     }
 }
--- a/tests/simpleharness/Program.cs
+++ b/tests/simpleharness/Program.cs
@@ -1,7 +1,10 @@
 using Microsoft.Xunit.Performance;
 using Microsoft.Xunit.Performance.Api;
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Xunit;
 
 [assembly: MeasureGCAllocations]
@@ -24,25 +27,29 @@ namespace simpleharness
         {
             var args = new string[] { "FFT", "LU", "MC", "MM", "SOR", "\u03C3", "x\u0305" };
             foreach (var arg in args)
-            {
                 yield return new object[] { new string[] { arg } };
-            }
         }
 
-        [Benchmark(InnerIterationCount = 10000)]
+        [Benchmark(InnerIterationCount = 10)]
         [MemberData(nameof(InputData))]
-        public static void TestBenchmark(string[] args)
+        private static void TestMultipleStringInputs(string[] args)
         {
-            foreach (BenchmarkIteration iter in Benchmark.Iterations)
-            {
-                using (iter.StartMeasurement())
-                {
-                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
-                    {
-                        string.Format("{0}{1}{2}{3}", args[0], args[0], args[0], args[0]);
-                    }
-                }
-            }
+           foreach (BenchmarkIteration iter in Benchmark.Iterations)
+           {
+               using (iter.StartMeasurement())
+               {
+                   for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                   {
+                       FormattedString(args[0], args[0], args[0], args[0]);
+                   }
+               }
+           }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static string FormattedString(string a, string b, string c, string d)
+        {
+            return string.Format("{0}{1}{2}{3}", a, b, c, d);
         }
 
         public sealed class Type_1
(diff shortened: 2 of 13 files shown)


src/xunit.performance.api/BasePerformanceMetric.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
+using Microsoft.Xunit.Performance.Sdk;
+using System.Collections.Generic;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal abstract class BasePerformanceMetric : PerformanceMetric
```
brianrob: I would make this base class more specific to PMCs, because it's all PMC specific.  You would not be able to use this class for an arbitrary ETW metric.

src/xunit.performance.api/GenericPerformanceMetric.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Microsoft.Xunit.Performance.Sdk;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal sealed class GenericPerformanceMetric<T> : BasePerformanceMetric
```
brianrob: Same here.  You should make this PMC specific.

src/xunit.performance.api/GenericPerformanceMetricDiscoverer.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.Diagnostics.Tracing.Session;
+using Microsoft.Xunit.Performance.Sdk;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit.Abstractions;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal sealed class GenericPerformanceMetricDiscoverer<T> : IPerformanceMetricDiscoverer
+        where T : IPerformanceMonitorCounter, new()
```
brianrob: Same here.  You should make this PMC specific.

src/xunit.performance.api/GenericPerformanceMetricEvaluator.cs
```diff
@@ -0,0 +1,57 @@
...
+        where T : IPerformanceMonitorCounter, new()
+    {
+        private PerformanceMetricEvaluationContext _context;
+        private int _profileSource;
+        private int _interval;
+        private long _count;
+
+        internal void Initialize(PerformanceMetricEvaluationContext context, int profileSourceInfoID)
+        {
+            lock (this)
```
brianrob: Do you need this lock?

src/xunit.performance.api/GenericPerformanceMetricEvaluator.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Parsers.Kernel;
+using Microsoft.Xunit.Performance.Sdk;
+using System.Diagnostics;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal class GenericPerformanceMetricEvaluator<T> : PerformanceMetricEvaluator
+        where T : IPerformanceMonitorCounter, new()
```
brianrob: Same here.  You should make this PMC specific.

src/xunit.performance.api/XunitBenchmark.cs
```diff
@@ -52,6 +54,15 @@ static XunitBenchmark()
                     ),
                 }
             };
+
+            foreach (var pmc in PerformanceMonitorCounters)
```
brianrob: If someone wants to add a new PMC metric that is not captured by default, how do they get their data captured without it being in the list of PerformanceMonitorCounters (meaning that this is just here because we are forcing it on)?

src/xunit.performance.api/XunitBenchmark.cs
```diff
@@ -78,11 +89,26 @@ public static (IEnumerable<ProviderInfo> providers, IEnumerable<PerformanceTestM
                         select provider;
 
                     providers = ProviderInfo.Merge(providers);
+
+                    // Inject implicit pmc counters
+                    foreach (var test in testMessageVisitor.Tests)
+                    {
+                        var metrics = new List<PerformanceMetricInfo>(test.Metrics);
+                        foreach (var performanceMetricInfo in PerformanceMetricInfos)
+                            if (performanceMetricInfo.IsValidPmc)
```
brianrob: Is IsValidPmc now part of the base class for metricInfo?

src/xunit.performance.api/GenericPerformanceMetricEvaluator.cs
jorive: This lock was a port from the Instructions Retired class.

src/xunit.performance.api/BasePerformanceMetric.cs
jorive: More PMC specific == Rename the class?

src/xunit.performance.api/GenericPerformanceMetric.cs
jorive: same as above?

src/xunit.performance.api/GenericPerformanceMetricDiscoverer.cs
jorive: same as above?

src/xunit.performance.api/XunitBenchmark.cs
jorive: Yes, we are forcing it on for the next perf review meeting to capture statistics that can help correlate data and benchmark noise.

src/xunit.performance.api/BasePerformanceMetric.cs
brianrob: Yes, please.

src/xunit.performance.api/GenericPerformanceMetric.cs
brianrob: Yup

src/xunit.performance.api/GenericPerformanceMetricDiscoverer.cs
brianrob: Yup.

src/xunit.performance.api/GenericPerformanceMetricEvaluator.cs
brianrob: Ok, sounds good.

src/xunit.performance.api/XunitBenchmark.cs
jorive: It is part of the base class for the pmc metrics.

src/xunit.performance.api/XunitBenchmark.cs
brianrob: Yup, no issue with forcing it on.  Just wanted to make sure that PMC sources can still be enabled via the existing mechanism.

src/xunit.performance.api/XunitBenchmark.cs
brianrob: Gotcha, so you're just iterating over the PMC metrics here.  Thanks.

jorive: That's right. This is a list of PMC, and I make sure I only add the ones that are available on the running platform.

src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMetric.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
+using Microsoft.Xunit.Performance.Sdk;
+using System.Collections.Generic;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal abstract class BasePerformanceMetric : PerformanceMetric
```
dpodder: Why does `BasePerformanceMetric` inherit from `PerformanceMetric`? Shouldn't it be the other way around? (Or is it just a naming issue?)

src/xunit.performance.api/XunitBenchmark.cs
```diff
@@ -14,23 +14,25 @@ static XunitBenchmark()
...
+            PerformanceMonitorCounters = new List<IPerformanceMonitorCounter>
+            {
+                new BranchMispredictionsPerformanceMonitorCounter(),
+                new CacheMissesPerformanceMonitorCounter(),
+            };
+
+            PerformanceMetricInfos = new List<BasePerformanceMetric>
+            {
+                new GenericPerformanceMonitorCounterMetric<BranchMispredictionsPerformanceMonitorCounter>(new BranchMispredictionsPerformanceMonitorCounter()),
+                new GenericPerformanceMonitorCounterMetric<CacheMissesPerformanceMonitorCounter>(new CacheMissesPerformanceMonitorCounter()),
```
dpodder: Do you need to explicitly spell out the generic argument type? Wouldn't the compiler deduce it from the type of the object being passed in as an argument?

src/xunit.performance.api/XunitBenchmark.cs
```diff
@@ -14,23 +14,25 @@ static XunitBenchmark()
...
             //   1. This reads as: flags read from config file based on running platform!
             //   2. Format could be something like this: `{ "Name" : "BranchMispredictions", "interval" : "100000" }`
-            RequiredProviders = new ProviderInfo[]
+            PerformanceMonitorCounters = new List<IPerformanceMonitorCounter>
+            {
+                new BranchMispredictionsPerformanceMonitorCounter(),
+                new CacheMissesPerformanceMonitorCounter(),
+            };
+
+            PerformanceMetricInfos = new List<BasePerformanceMetric>
```
dpodder: I'm guessing we need one `BasePerformanceMetric` for each `IPerformanceMonitorCounter` above, right? Could we somehow loop through the list and automatically create one, to avoid having two mirrored lists?

tests/simpleharness/Program.cs
```diff
@@ -65,35 +87,76 @@ public void TestBenchmark()
...
+                {
+                    var pos = rnd.Next(nCards);
+                    if (pos % 3 != 0)
+                        pos = rnd.Next(nCards);
+                    var temp = deck[i];
+                    deck[i] = deck[pos];
+                    deck[pos] = temp;
+                }
+
+//#if DEBUG
```
dpodder: deleteme?

src/xunit.performance.api/XunitBenchmark.cs
jorive: I get this otherwise:
`Using the generic type 'GenericPerformanceMonitorCounterMetric<T>' requires 1 type arguments`

dpodder: Darn, I guess not... oh well, in that case, let's leave it as-is.

src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMetric.cs
jorive: Renaming issue

src/xunit.performance.api/XunitBenchmark.cs
jorive: IPerformanceMonitorCounter is an interface that defines the configuration of the pmc to be used, not the internal of the pmc (at least for now). The idea is to have a user interface decoupled from the implementation of the pmc (the pmc id). Currently, this is hardcoded in the function, but you can imagine that this data is given by the API user instead (in the near future).

src/xunit.performance.api/XunitBenchmark.cs
jorive: This might work once I update the dotnet version (hint: next PR ;))

src/xunit.performance.api/XunitBenchmark.cs
dpodder: OK, particularly if this is to go away in the future, and since the lists only have two elements each, this looks reasonable.

src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMetric.cs
dpodder: Thanks, looks better now :smile:

dpodder:
:shipit:

PR #197, opened 2017-03-03
Remove CSwitch from the default collection

DrewScoggins:
By default we were collecting context switch events and not using them
during analysis.  This removes this because it greatly inflates the size
of the collected ETL file and was causing us to drop events and fail on
certain test.

### DIFF
--- a/src/xunit.performance.api/ETWProfiler.cs
+++ b/src/xunit.performance.api/ETWProfiler.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Xunit.Performance.Api
                     userEventSession.BufferSizeMB = bufferSizeMB;
 
                     var flags = KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.ImageLoad | KernelTraceEventParser.Keywords.Thread;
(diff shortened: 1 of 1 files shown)


DrewScoggins:
PTAL @jorive @brianrob

jorive:
By dropping this, we will not be collecting (Precise) CPU Usage, only (Sampled) CPU Usage.
Even though we are not collecting/uploading this data, it might be valuable for private investigations. Should we disable the failing tests instead? For what I understand, the test is questionably measuring something useful.

**Data collected by CPU Usage (Precise)**
- Utilization by process;
- Context switch count by process, thread;
- Context switch rate by CPU;
- Timeline by CPU;
- Timeline by process, thread;
- Usage by priority at context switch begin; utilization by CPU;
- Utilization by process, thread, stack

DrewScoggins:
It is true that the value of these tests is marginal, but it will be lower friction to check in the code for turning on runs in CoreFX if we do not also have to turn off a class of tests.  Also since we do not persist the ETL files if we ever needed context switch in order to calculate some metric we can add it back.  Since it is only a one line change it should be simple.

DrewScoggins:
We talked about this offline.  We will try other ways to fix this issue so closing this for now.

PR #203, opened 2017-03-03
Updating dotnet version and csproj files

jorive:
1. This change enable us to build/run with Visual Studio 2017.
2. Moved the download tools step to its own script.
3. Temporarily disabled creating packages using the nuspec files as the `dotnet nugget pack` scenario was removed, we are currently planning on moving old tools to their own branch, and we are not updating/uploading these packages.


dotnet-install.cmd
```diff
@@ -0,0 +1,45 @@
...
+  set Init_Tools_Log=%DotNet_Path%\install.log
+  set /p DotNet_Version=<"%~dp0DotNetCLIVersion.txt"
+  set DotNet_Installer_Url=https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1
+
+  REM dotnet.exe might exist, but it might not be the right version.
+  REM Here we verify that if it is not the right version, then we install it
+  if exist "%DotNet%" (
+    (call "%DotNet%" --version|findstr /i "%DotNet_Version%" 1>nul 2>&1) && goto :install_dotnet_cli_exit
+    call :print_error_message Current version of "%DotNet%" does not match expected version "%DotNet_Version%"
+    call :remove_directory "%DotNet_Path%" || exit /b 1
```
dpodder: I think these `:subroutines` are missing in this cmd

dotnet-install.cmd
```diff
@@ -0,0 +1,45 @@
...
+:install_dotnet_cli
+setlocal
+  set DotNet_Path=%~dp0tools\bin
+  set Init_Tools_Log=%DotNet_Path%\install.log
+  set /p DotNet_Version=<"%~dp0DotNetCLIVersion.txt"
+  set DotNet_Installer_Url=https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1
+
+  REM dotnet.exe might exist, but it might not be the right version.
+  REM Here we verify that if it is not the right version, then we install it
+  if exist "%DotNet%" (
```
dpodder: I'm guessing `%DotNet%` has to be defined by the caller? Could we define it here to make this script less prone to break?

jorive: I am thinking it should be on the PATH, based on the current dotnet version (dotnet install script should add it).

dotnet-install.cmd
jorive: Fixing it :)

PR #204, opened 2017-03-06
Adding allocations for current thread, bug fixes, and stop building old runner.

jorive:
1. Updated how small decimal numbers are displayed on the statistics
2. Started initial work to gather the allocated bytes in the current thread (this work on .NET Core 1.1)
3. Trimmed the markdown string, which was not being rendered as a table on GitHub
4. Updated build script (disabled old runner from building from the script and made the api the main and only project to be built)

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -10,9 +10,6 @@ setlocal enabledelayedexpansion
   set VersionSuffix=%~2
   if "%VersionSuffix%"=="" set VersionSuffix=beta-build0000
 
-  set PackageVersion=%~3
-  if "%PackageVersion%"=="" set PackageVersion=1.0.0-%VersionSuffix%
-
   set OutputDirectory=%~dp0LocalPackages
   call :remove_directory "%OutputDirectory%" || exit /b 1
 
--- a/build.cmd
+++ b/build.cmd
@@ -4,13 +4,6 @@
 :main
 setlocal enabledelayedexpansion
   set errorlevel=
-
-  set lv_api_only=
-  if /I "%~1" == "--api-only" (
-    set lv_api_only=1
-    shift
-  )
-
   set BuildConfiguration=%~1
   if "%BuildConfiguration%"=="" set BuildConfiguration=Debug
 
@@ -26,26 +19,9 @@ setlocal enabledelayedexpansion
   call "%~dp0.\dotnet-install.cmd" || exit /b 1
 
   set procedures=
-
-  if not defined lv_api_only (
-    set procedures=!procedures! build_procdomain
-    set procedures=!procedures! build_xunit_performance_analysis
-    set procedures=!procedures! build_xunit_performance_core
-    set procedures=!procedures! build_xunit_performance_execution
-    set procedures=!procedures! build_xunit_performance_metrics
-    set procedures=!procedures! build_xunit_performance_logger
-    set procedures=!procedures! build_xunit_performance_run
-    set procedures=!procedures! build_microsoft_dotnet_xunit_performance_runner_cli
-    set procedures=!procedures! build_microsoft_dotnet_xunit_performance_analysis_cli
-    set procedures=!procedures! build_samples_classlibrary_net46
-    set procedures=!procedures! build_samples_simpleperftests
-    REM set procedures=!procedures! nuget_pack_src
-  ) else (
-    set procedures=!procedures! build_xunit_performance_core
-    set procedures=!procedures! build_xunit_performance_execution
-    set procedures=!procedures! build_xunit_performance_metrics
-  )
-
+  set procedures=%procedures% build_xunit_performance_core
+  set procedures=%procedures% build_xunit_performance_execution
+  set procedures=%procedures% build_xunit_performance_metrics
   set procedures=%procedures% build_xunit_performance_api
   set procedures=%procedures% build_tests_simpleharness
 
@@ -55,19 +31,7 @@ setlocal enabledelayedexpansion
       exit /b 1
     )
   )
-  exit /b %errorlevel%
-
-:build_procdomain
-setlocal
-  cd /d %~dp0src\procdomain
-  call :dotnet_build
-  exit /b %errorlevel%
-
-:build_xunit_performance_analysis
-setlocal
-  cd /d %~dp0src\xunit.performance.analysis
-  call :dotnet_build
-  exit /b %errorlevel%
+endlocal& exit /b %errorlevel%
 
 :build_xunit_performance_core
 setlocal
@@ -81,57 +45,11 @@ setlocal
   call :dotnet_pack
   exit /b %errorlevel%
 
-:build_xunit_performance_logger
-setlocal
-  cd /d %~dp0src\xunit.performance.logger
-  call :dotnet_build
-  exit /b %errorlevel%
-
 :build_xunit_performance_metrics
 setlocal
   cd /d %~dp0src\xunit.performance.metrics
   call :dotnet_pack
   exit /b %errorlevel%
-
-:build_xunit_performance_run
-setlocal
-  cd /d %~dp0src\xunit.performance.run
-  call :dotnet_build
-  exit /b %errorlevel%
-
-:build_microsoft_dotnet_xunit_performance_runner_cli
-setlocal
-  cd /d %~dp0src\cli\Microsoft.DotNet.xunit.performance.runner.cli
-  call :dotnet_build
-  exit /b %errorlevel%
-
-:build_microsoft_dotnet_xunit_performance_analysis_cli
-setlocal
-  cd /d %~dp0src\cli\Microsoft.DotNet.xunit.performance.analysis.cli
-  call :dotnet_build
-  exit /b %errorlevel%
-
-:build_samples_classlibrary_net46
-setlocal
-  cd /d %~dp0samples\ClassLibrary.net46
-  call :dotnet_build
-  exit /b %errorlevel%
-
-:build_samples_simpleperftests
-setlocal
-  cd /d %~dp0samples\SimplePerfTests
-  call :dotnet_build
-  exit /b %errorlevel%
-
-:nuget_pack_src
-setlocal
-  cd /d %~dp0src
-  dotnet.exe restore xunit.performance.csproj                                                                                                                           || exit /b 1
-  dotnet.exe pack xunit.performance.csproj                 --no-build -c %BuildConfiguration% -o "%OutputDirectory%" --version-suffix %VersionSuffix% --include-symbols || exit /b 1
-
-  dotnet.exe restore xunit.performance.runner.Windows.csproj                                                                                                            || exit /b 1
-  dotnet.exe pack xunit.performance.runner.Windows.csproj  --no-build -c %BuildConfiguration% -o "%OutputDirectory%" --version-suffix %VersionSuffix% --include-symbols || exit /b 1
-  exit /b 0
 
 :build_xunit_performance_api
 setlocal
--- a/build.cmd
+++ b/build.cmd
@@ -8,7 +8,7 @@ setlocal enabledelayedexpansion
   if "%BuildConfiguration%"=="" set BuildConfiguration=Debug
 
   set VersionSuffix=%~2
-  if "%VersionSuffix%"=="" set VersionSuffix=alpha-build0000
+  if "%VersionSuffix%"=="" set VersionSuffix=beta-build0000
 
   set PackageVersion=%~3
   if "%PackageVersion%"=="" set PackageVersion=1.0.0-%VersionSuffix%
@@ -65,7 +65,13 @@ setlocal
     call :print_error_message Cannot run simpleharness test because this is not an administrator window
     exit /b 1
   )
-  dotnet.exe run -c %BuildConfiguration% "bin\%BuildConfiguration%\netcoreapp1.0\simpleharness.dll" || exit /b 1
+
+  dotnet.exe publish -c %BuildConfiguration% --framework netcoreapp1.0  || exit /b 1
+  dotnet.exe publish -c %BuildConfiguration% --framework netcoreapp1.1  || exit /b 1
+
+  rem Use this as a prefix for dotnet.exe in order to launch the process on a different window: START "TITLE GOES HERE" /WAIT
+  dotnet.exe "bin\%BuildConfiguration%\netcoreapp1.0\simpleharness.dll" || exit /b 1
+  dotnet.exe "bin\%BuildConfiguration%\netcoreapp1.1\simpleharness.dll" || exit /b 1
   exit /b %errorlevel%
 
 :dotnet_build
--- a/src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
+++ b/src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
@@ -8,7 +8,6 @@ using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftXunitBenchmark;
 using Microsoft.Xunit.Performance.Sdk;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.Tracing;
 using System.Linq;
 
 namespace Microsoft.Xunit.Performance.Api
--- a/src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
+++ b/src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
@@ -8,6 +8,7 @@ using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftXunitBenchmark;
 using Microsoft.Xunit.Performance.Sdk;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
 using System.Linq;
 
 namespace Microsoft.Xunit.Performance.Api
--- a/src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
+++ b/src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
@@ -7,10 +7,8 @@ namespace Microsoft.Xunit.Performance.Api
 {
     internal class GCAllocatedBytesForCurrentThreadEvaluator : PerformanceMetricEvaluator
     {
-        public GCAllocatedBytesForCurrentThreadEvaluator(PerformanceMetricEvaluationContext context)
+        public GCAllocatedBytesForCurrentThreadEvaluator()
         {
-            _context = context;
-            //_context.TraceEventSource.UserData
         }
 
         public override void BeginIteration(TraceEvent traceEvent)
@@ -25,7 +23,6 @@ namespace Microsoft.Xunit.Performance.Api
                 _bytesBeforeIteration, bytesAfterIteration);
         }
 
-        private readonly PerformanceMetricEvaluationContext _context;
         private long _bytesBeforeIteration;
     }
 }
--- a/src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
+++ b/src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
@@ -1,28 +1,18 @@
 ﻿using Microsoft.Diagnostics.Tracing;
 using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftXunitBenchmark;
-using Microsoft.Xunit.Performance.Execution;
 using Microsoft.Xunit.Performance.Sdk;
 
 namespace Microsoft.Xunit.Performance.Api
 {
     internal class GCAllocatedBytesForCurrentThreadEvaluator : PerformanceMetricEvaluator
     {
-        public GCAllocatedBytesForCurrentThreadEvaluator()
-        {
-        }
-
         public override void BeginIteration(TraceEvent traceEvent)
         {
-            _bytesBeforeIteration = ((BenchmarkIterationStartArgs)traceEvent).AllocatedBytes;
         }
 
         public override double EndIteration(TraceEvent traceEvent)
         {
-            var bytesAfterIteration = ((BenchmarkIterationStopArgs)traceEvent).AllocatedBytes;
-            return AllocatedBytesForCurrentThread.GetTotalAllocatedBytes(
-                _bytesBeforeIteration, bytesAfterIteration);
+            return ((BenchmarkIterationStopArgs)traceEvent).AllocatedBytes;
         }
-
-        private long _bytesBeforeIteration;
     }
 }
--- a/src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
+++ b/src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
@@ -1,0 +1,31 @@
+﻿using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftXunitBenchmark;
+using Microsoft.Xunit.Performance.Execution;
+using Microsoft.Xunit.Performance.Sdk;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal class GCAllocatedBytesForCurrentThreadEvaluator : PerformanceMetricEvaluator
+    {
+        public GCAllocatedBytesForCurrentThreadEvaluator(PerformanceMetricEvaluationContext context)
+        {
+            _context = context;
+            //_context.TraceEventSource.UserData
+        }
+
+        public override void BeginIteration(TraceEvent traceEvent)
+        {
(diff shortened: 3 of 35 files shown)


src/xunit.performance.api/XunitRunner.cs
```diff
@@ -70,7 +72,25 @@ private static void SetupRunnerCallbacks(AssemblyRunner runner, ManualResetEvent
...
                 lock (consoleLock)
                 {
-                    WriteInfoLine($"Running {info.TestCasesToRun} Benchmarks out of {info.TestCasesDiscovered} Xunit Facts...");
+                    var diff = info.TestCasesDiscovered - info.TestCasesToRun;
+                    if (diff < 0) // This should never happen.
+                    {
+                        // TODO: Add error handling. Throwing exception is not enough because we are waiting for the async runner to finish.
+                        WriteErrorLine("There are more benchmarks than facts.");
+                    }
+                    var message = (diff == 0) ? $"Running {info.TestCasesToRun} Benchmarks..." : $"Running {info.TestCasesToRun} Benchmarks out of {info.TestCasesDiscovered} Xunit Facts...";
```
dpodder: I think this would be easier to interpret for the `diff != 0` case:
`$"Running {info.TestCasesToRun} Benchmarks, skipping {diff} Xunit Facts..."`

tests/simpleharness/simpleharness.csproj
```diff
@@ -6,7 +6,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
```
jorive: @brianrob @DrewScoggins Would you know how to check this version at runtime?

src/xunit.performance.execution/xunit.performance.execution.csproj
```diff
@@ -7,7 +7,7 @@
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.execution</AssemblyTitle>
     <Description>xUnit Performance Execution</Description>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netstandard1.5</TargetFramework>
```
dpodder: Does this mean we're dropping support for 1.3? Or has this been the case from the start?

src/xunit.performance.metrics/xunit.performance.metrics.csproj
```diff
@@ -7,7 +7,7 @@
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.metrics</AssemblyTitle>
     <Description>xUnit Performance Metrics</Description>
-    <TargetFrameworks>netstandard1.5;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard1.5</TargetFrameworks>
```
dpodder: Should this drop the trailing `s` and become just `<TargetFramework>` now?

src/xunit.performance.execution/xunit.performance.execution.csproj
jorive: It has been the case from the start. People using the API, use the API package directly, which is 1.5

src/xunit.performance.metrics/xunit.performance.metrics.csproj
jorive: Yes, we can.

src/xunit.performance.api/EtwPerformanceMetricEvaluationContext.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Xunit.Performance.Sdk;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
```
jorive: Good catch! :)

src/xunit.performance.execution/xunit.performance.execution.csproj
dpodder: OK, then it's fine. Just wanted to make sure. :smile:

src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public override void BeginIteration(TraceEvent traceEvent)
+        {
+            _bytesBeforeIteration = ((BenchmarkIterationStartArgs)traceEvent).AllocatedBytes;
+        }
+
+        public override double EndIteration(TraceEvent traceEvent)
+        {
+            var bytesAfterIteration = ((BenchmarkIterationStopArgs)traceEvent).AllocatedBytes;
```
brianrob: Rather than doing the calculation here, I think the calculation should be done before the event is written.  The written Iteration/Stop event should contain the final number so that as the harness evolves and changes are made (or the runtime stops allocating extra objects), re-processing of traces is not broken because the overhead of this feature changes.

I expect us to want to build the "re-process my trace with this new analysis" feature at some point in the future.

src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Xunit.Performance.Sdk;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal class GCAllocatedBytesForCurrentThreadEvaluator : PerformanceMetricEvaluator
+    {
+        public GCAllocatedBytesForCurrentThreadEvaluator(PerformanceMetricEvaluationContext context)
+        {
+            _context = context;
+            //_context.TraceEventSource.UserData
```
brianrob: Extra comment?

src/xunit.performance.execution/AllocatedBytesForCurrentThread.cs
```diff
@@ -0,0 +1,48 @@
...
+            var allocatedBytesBefore = LastAllocatedBytes;
+            var allocatedBytesAfter = LastAllocatedBytes;
+            s_minAllocatedBytes = allocatedBytesAfter - allocatedBytesBefore;
+        }
+
+        public static long LastAllocatedBytes => _GetAllocatedBytesForCurrentThread.Invoke();
+
+        public static long GetTotalAllocatedBytes(long allocatedBytesBefore, long allocatedBytesAfter)
+        {
+            return allocatedBytesAfter - allocatedBytesBefore; // FIXME: (allocatedBytesAfter - allocatedBytesBefore - s_minAllocatedBytes) sometimes returns a negative value!
```
brianrob: I think we need to investigate and fix this issue before merging.  Otherwise, we are going to have incorrect values being returned.  Per our offline conversation, I expect a trace with benchmark iteration events and all allocations should help us get to the bottom of this.

src/xunit.performance.execution/BenchmarkEventSource.cs
```diff
@@ -121,23 +115,27 @@ public unsafe void BenchmarkStop(string RunId, string BenchmarkName, string Stop
...
                 }
             }
 
             if (_csvWriter != null)
                 WriteCSV(BenchmarkName, stopReason: StopReason);
         }
 
-        [Event(3, Opcode = EventOpcode.Info, Task = Tasks.BenchmarkIterationStart)]
-        public unsafe void BenchmarkIterationStart(string RunId, string BenchmarkName, int Iteration)
+        [Event(3,
```
brianrob: Nit: Can you put this attribute all on one line like the rest?

src/xunit.performance.execution/BenchmarkEventSource.cs
```diff
@@ -121,23 +115,27 @@ public unsafe void BenchmarkStop(string RunId, string BenchmarkName, string Stop
...
+            Level = EventLevel.LogAlways,
```
brianrob: Do you need to specify Level?

src/xunit.performance.execution/BenchmarkEventSource.cs
```diff
@@ -121,23 +115,27 @@ public unsafe void BenchmarkStop(string RunId, string BenchmarkName, string Stop
...
+            Opcode = EventOpcode.Start,
+            Task = Tasks.BenchmarkIteration)]
+        public unsafe void BenchmarkIterationStart(string RunId, string BenchmarkName, int Iteration, long AllocatedBytes)
         {
             if (_csvWriter != null)
                 WriteCSV(BenchmarkName, iteration: Iteration);
```
brianrob: For Linux, you need to specify the allocated bytes to the CSV here, right?

src/xunit.performance.execution/BenchmarkEventSource.cs
```diff
@@ -147,45 +145,61 @@ public unsafe void BenchmarkIterationStart(string RunId, string BenchmarkName, i
...
+                    data[3].DataPointer = (IntPtr)(&AllocatedBytes);
+
+                    WriteEventCore(3, eventDataCount, data);
                 }
             }
         }
 
-        [Event(4, Opcode = EventOpcode.Info, Task = Tasks.BenchmarkIterationStop)]
-        public unsafe void BenchmarkIterationStop(string RunId, string BenchmarkName, int Iteration)
+        [Event(4,
```
brianrob: Nit: Can you put this all on one line?

Also, do you need to specify Level? It is not specified for the other events and wasn't specified for this event until this PR.

src/xunit.performance.execution/BenchmarkTestInvoker.cs
```diff
@@ -190,7 +190,7 @@ protected override void StartMeasurement(int iterationNumber)
 
                     RandomizeMeasurementStartTime();
 
-                    BenchmarkEventSource.Log.BenchmarkIterationStart(BenchmarkConfiguration.Instance.RunId, _testName, iterationNumber);
+                    BenchmarkEventSource.Log.BenchmarkIterationStart(BenchmarkConfiguration.Instance.RunId, _testName, iterationNumber, 0);
```
brianrob: With my suggestion further up in the review to calculate the actual allocated bytes before logging, you should not need this change because the iteration/stop event is the only one that will change.

src/xunit.performance.execution/BenchmarkTestInvoker.cs
```diff
@@ -252,7 +252,7 @@ protected override void StopMeasurement(int iterationNumber)
                     Debug.Assert(_currentIterationMeasurementStarted);
                     _currentIterationMesaurementStopped = true;
 
-                    BenchmarkEventSource.Log.BenchmarkIterationStop(BenchmarkConfiguration.Instance.RunId, _testName, iterationNumber);
+                    BenchmarkEventSource.Log.BenchmarkIterationStop(BenchmarkConfiguration.Instance.RunId, _testName, iterationNumber, 0);
```
brianrob: You should calculate allocated bytes here before calling BenchmarkIterationStop and then pass it in here.  Then everything works out with traceparsergen and you can do whatever calculation you need outside of the eventsource (it's really just for logging).

src/xunit.performance.metrics/Microsoft-Xunit-Benchmark.manifest.cs
```diff
@@ -205,7 +209,7 @@ public sealed class BenchmarkIterationStopArgs : TraceEvent
         public string RunId { get { return GetUnicodeStringAt(0); } }
         public string BenchmarkName { get { return GetUnicodeStringAt(SkipUnicodeString(0)); } }
         public int Iteration { get { return GetInt32At(SkipUnicodeString(SkipUnicodeString(0))); } }
-        public bool Success { get { return GetInt32At(SkipUnicodeString(SkipUnicodeString(0)) + 4) != 0; } }
+        public long AllocatedBytes { get { return GetInt64At(SkipUnicodeString(SkipUnicodeString(0)) + 4); } }
```
brianrob: Should not need this in Iteration/Start.

tests/simpleharness/simpleharness.csproj
brianrob: I'm not sure that there is a way to do this.  What do you need to do this for?

build.cmd
```diff
@@ -26,26 +19,9 @@ setlocal enabledelayedexpansion
   call "%~dp0.\dotnet-install.cmd" || exit /b 1
 
   set procedures=
-
-  if not defined lv_api_only (
-    set procedures=!procedures! build_procdomain
```
brianrob: Before removing all of this, I think we need two things:

1. Update the README to cover how the harness works.
2. Write and reference an issue from this PR that describes what's happening and why.

It's fine to include this change in this PR, but we just need to do these things before merging.

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -70,7 +72,26 @@ private static void SetupRunnerCallbacks(AssemblyRunner runner, ManualResetEvent
...
+                        WriteInfoLine($"Skipping {diff} Xunit [Fact]s because they are not [Benchmark]s");
+                }
+            };
+            runner.OnErrorMessage = info =>
+            {
+                lock (consoleLock)
+                {
+                    var sb = new StringBuilder();
+                    sb.AppendLine("Got an unhandled exception outside test.");
+                    sb.AppendLine($"  {info.ExceptionType}: {info.MesssageType}: {info.ExceptionMessage}");
```
jorive: This is an `ErrorMessageInfo` object from Xunit, not an exception.

src/xunit.performance.api/xunit.performance.api.csproj
```diff
@@ -20,16 +20,13 @@
     <Compile Include="..\common\GlobalAssemblyInfo.cs" Link="GlobalAssemblyInfo.cs" />
   </ItemGroup>
 
-  <PropertyGroup>
-    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
```
jorive: It is inferred from here: `<TargetFramework>netstandard1.5</TargetFramework>`.
I was able to cleanup the csproj, and I'm not building the old project anymore. Thus, I was able to remove building net46 and some warnings we used to get.

src/xunit.performance.execution/AllocatedBytesForCurrentThread.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </remarks>
+        static AllocatedBytesForCurrentThread()
+        {
+            var typeInfo = typeof(GC).GetTypeInfo();
+            var method = typeInfo.GetMethod("GetAllocatedBytesForCurrentThread",
+                            BindingFlags.Public | BindingFlags.Static)
+                      ?? typeInfo.GetMethod("GetAllocatedBytesForCurrentThread",
+                            BindingFlags.NonPublic | BindingFlags.Static);
+
+            // TODO: Add meaningful error message.
```
jorive: I think we should at least output a `warning`, unable to capture this metric on netcoreapp with version less than 1.1. What do you think?

src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadEvaluator.cs
jorive: Context is not used anymore. I'll delete this.

src/xunit.performance.execution/BenchmarkEventSource.cs
jorive: Yes, I will need to, but I was not planning to do it in this PR.

tests/simpleharness/simpleharness.csproj
jorive: I would like to print "warning" message (from the API), so users will know that this metric will not be captured.

brianrob: I see.  I think rather than detecting version, you should detect whether or not the API is present and then show a warning.  That way, you don't have to worry about what version was compiled against or someone inserting privates, etc.

src/xunit.performance.api/XunitRunner.cs
brianrob: Gotcha.

src/xunit.performance.api/xunit.performance.api.csproj
brianrob: Cool

src/xunit.performance.execution/AllocatedBytesForCurrentThread.cs
brianrob: I think that's a reasonable approach.  Will you add this as part of this PR?

tests/simpleharness/simpleharness.csproj
jorive: Makes sense.

src/xunit.performance.execution/AllocatedBytesForCurrentThread.cs
jorive: Yes, I will add it as part of this PR.

src/xunit.performance.execution/BenchmarkEventSource.cs
brianrob: I think you probably should.  I suspect that the first thing that will happen after you add this functionality for Windows is that it will be tried on Linux and won't work.  It would be good - unless there is some architectural issue - to complete these types of features for all platforms at once.

src/xunit.performance.api/XunitRunner.cs
```diff
@@ -70,7 +72,26 @@ private static void SetupRunnerCallbacks(AssemblyRunner runner, ManualResetEvent
             {
                 lock (consoleLock)
                 {
-                    WriteInfoLine($"Running {info.TestCasesToRun} Benchmarks out of {info.TestCasesDiscovered} Xunit Facts...");
+                    var diff = info.TestCasesDiscovered - info.TestCasesToRun;
+                    if (diff < 0) // This should never happen.
+                    {
+                        // TODO: Add error handling. Throwing exception is not enough because we are waiting for the async runner to finish.
```
jorive: https://github.com/Microsoft/xunit-performance/issues/205

build.cmd
jorive: Opened #206

jorive: @brianrob @dpodder Could you please take a look? I have updated all sources based on feedback, added some error checking, bug fixes, updated readme, remove runner projects and disabled the Linux build script (there is not CLI to be built, and I would need to update the groovy file first, in future PR)

build.cmd
```diff
@@ -4,48 +4,21 @@
...
 
   call "%~dp0.\dotnet-install.cmd" || exit /b 1
 
   set procedures=
-
-  if not defined lv_api_only (
-    set procedures=!procedures! build_procdomain
-    set procedures=!procedures! build_xunit_performance_analysis
-    set procedures=!procedures! build_xunit_performance_core
-    set procedures=!procedures! build_xunit_performance_execution
```
brianrob: You should modify the name of your PR to make sure that it's clear that there is a big change here where you no longer build the runner.

src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadMetric.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Xunit.Performance.Sdk;
+using System.Collections.Generic;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    internal class GCAllocatedBytesForCurrentThreadMetric : PerformanceMetric
+    {
+        public GCAllocatedBytesForCurrentThreadMetric()
+            : base("GCAllocatedBytesForCurrentThread", "Allocations In Current Thread", PerformanceMetricUnits.Bytes)
```
brianrob: I think we may want to name the metric differently, since I'm not sure it will be clear to the viewer of data what "Current Thread" is.  Maybe "Allocation Size on Benchmark Execution Thread" or something like that....

src/xunit.performance.api/XunitBenchmark.cs
```diff
@@ -90,16 +91,24 @@ public static (IEnumerable<ProviderInfo> providers, IEnumerable<PerformanceTestM
...
+                        defaultMetricInfos = defaultMetricInfos.Concat(new[] { new GCAllocatedBytesForCurrentThreadMetric() });
+                    else
                     {
-                        var metrics = new List<PerformanceMetricInfo>(test.Metrics);
-                        foreach (var performanceMetricInfo in PerformanceMetricInfos)
-                            if (performanceMetricInfo.IsValidPmc)
-                                metrics.Add(performanceMetricInfo);
-                        test.Metrics = metrics;
+                        WriteWarningLine(AllocatedBytesForCurrentThread.NoAvailabilityReason);
+                        WriteWarningLine("The 'Allocations In Current Thread' metric will not be collected.");
```
brianrob: Name change here as well... I think you should probably get the name from the metric here so that it's always consistent.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -104,6 +105,10 @@ public static XunitPerformanceHarnessOptions Parse(string[] args)
                                     Console.WriteLine(Usage());
                                     Environment.Exit(0);
                                     break;
+                                case ErrorType.VersionRequestedError:
```
brianrob: Can you tell me more about the expectation here?  When would this error be printed?  It looks like right now, the error is just a version number, but doesn't say what the error is.

src/xunit.performance.execution/AllocatedBytesForCurrentThread.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                _GetAllocatedBytesForCurrentThread = () => (long)method.Invoke(null, null);
+                if (IsBugFixed())
+                {
+                    IsAvailable = true;
+                    NoAvailabilityReason = "";
+                }
+                else
+                {
+                    IsAvailable = false;
```
brianrob: I think for the second case where you encounter the bug - you can point at the issue on GH (show an actual URL) and just say "Not capturing <metric name> because the targeted runtime does not contain the fix for http://github.com/path/to/issue".  Then, it's super clear to users who care about this what to do.

tests/simpleharness/GetAllocatedBytesForCurrentThreadTest.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace simpleharness
+{
+    public class GetAllocatedBytesForCurrentThreadTest
+    {
+        [Benchmark]
+        public static void WithCollect()
+        {
+            var method = typeof(GC)
+                .GetTypeInfo()
```
brianrob: I think it would be good to have a project of functional tests that we run as well that run as part of the build.  These should live there.  It's good that they are [Benchmark]s and not [Fact]s, because we'll test our code, but I think that simpleharness should just be a simple example, and we should put all of these great test cases that you're building into a functional tests project that we run on every build.  Then devs will see breaks from their code, and the CI will automatically run them as well.

src/xunit.performance.metrics/Microsoft-Xunit-Benchmark.manifest.cs
jorive: Removed

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
jorive: This is executed when the user passes --version to the Harness constructor (request the version of the harness). e.g. `dotnet.exe --version` outputs 1.0.0

tests/simpleharness/GetAllocatedBytesForCurrentThreadTest.cs
jorive: The `tests/standalonesample` is the example of a simpleharness. The latter is a project with direct project references to the API (a non-user scenario).
I will be happy to clean this up on subsequent PR =)

src/xunit.performance.execution/AllocatedBytesForCurrentThread.cs
jorive: fixed

src/xunit.performance.api/XunitBenchmark.cs
jorive: fixed

src/xunit.performance.api/PerformanceMetric/GCAllocatedBytesForCurrentThreadMetric.cs
jorive: Fixed

build.cmd
jorive: fixed

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
brianrob: Got it.

tests/simpleharness/GetAllocatedBytesForCurrentThreadTest.cs
brianrob: Yup, we talked about this offline as well.  Sounds good.  I think we should fix it up in a subsequent PR so that we have "examples" and "tests".

brianrob:
You should squash your changes to a logical set of commits.  Otherwise, LGTM.  Thanks for making these changes.

Issue #202, opened 2017-03-03
Add Ability to Run Larger Benchmarks


brianrob:
Currently, xunit.performance is good at running microbenchmarks in-proc, but doesn't have great support for larger scenario-level benchmarks.

brianrob:
This is currently being worked on by @guhuro.

brianrob:
Fixed by #208.

PR #208, opened 2017-03-20
Add ability to run larger benchmarks

guhuro:
We add new functionality to the xunitPerformanceApi to support custom
benchmarks. The function RunStartupBenchmark receives arguments an three
delegates Pre, Mid and Post to be called before all the runs, after each
one and after finishing.
Via this delegates, we let the test authors to produce and collect the
measuring for its runs however they want.

We add a new class PerformanceTestConfig to collect all the
configuration needed for launching this tests. It's been created
thinking in JitBench, but it can be modified to include more options and
setups.


guhuro:
Addressing this issue https://github.com/Microsoft/xunit-performance/issues/202 
Can you take a look? 
@DrewScoggins @brianrob @jorive @dpodder @adiaaida

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +175,143 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
 
     [Serializable]
+    [XmlRoot("BenchmarkScenario")]
+    public sealed class BenchmarkScenario
+    {
+        [XmlAttribute("Name")]
+        public string Name { get; set; }
+
+        [XmlAttribute("Namespace")]
+        public string Namespace { get; set; }
```
brianrob: It looks like this is always empty string.  Do we need it?

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +175,143 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
+        
+        [XmlArray("Assemblies")]
+        public List<NetcoreAssembly> Assemblies { get; set; }
```
brianrob: Do scenarios need an assembly list?  Since they have no coupling to xunit or any other assembly based runner, I wouldn't expect assemblies to be tracked here.

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +175,143 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
+
+        public BenchmarkScenario()
+        {
+            Assemblies = new List<NetcoreAssembly>();
```
brianrob: If we do need to track assemblies, we should come up with a different name for this class, because the harness can run on more than just .NET core, so the assemblies are not specific to .NET core.

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +175,143 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
+                    new XmlSerializer(typeof(BenchmarkScenario))
+                        .Serialize(sw, this, namespaces);
+                }
+            }
+        }
+
+        internal DataTable GetStatistics()
+        {
+            var dt = new DataTable();
+            var col0_testName = dt.AddColumn("Test Name");
```
brianrob: Are there constants for these column names?  I think these are duplicated from the xunit runner code, so we should probably have constants for these to avoid breaks when a name is changed.

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +175,143 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
+        public NetcoreAssembly(string name)
+        {
+            Name = name;
+            Tests = new List<NetcoreAssemblyTest>();
+        }
+    }
+
+    [Serializable]
+    [XmlType("Test")]
+    public sealed class NetcoreAssemblyTest
```
brianrob: I think we probably want to call this class "TestModel" or something like that - the XML type can remain as is.  Same thing here - nothing .NET core specific.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
             s_profiler(assemblyPath, Configuration.RunId, _outputDirectory, runner, collectOutputFilesCallback);
         }
 
+        public void RunStartupBenchmark(string[] args, ProcessStartInfo psi, PreDel pre, MidDel mid, PostDel post)
```
brianrob: A few comments for this API:

1. This is not actually doing a start-up benchmark.  JitBench happens to do that, but the API you've designed is fairly high-level and can support any number of metrics and arbitrary tests - not just start-up.  I'd recommend calling this something like "RunScenario", but I'm open to other names.
2. The delegate names should be expanded to describe exactly what they're for.  PreDel should probably be called something like "SetupDelegate.  PostDel should be something like "TeardownDelegate".  MidDel should probably be something like "PostIterationDelegate".
3. The argument names should be more fully worded so that someone looking at them can know what they are for.
4. We should clarify how one knows that the run succeeded.  I'm assuming that it's "we didn't get an exception", but want to be sure.

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,18 @@
+
+
+
+namespace Microsoft.Xunit.Performance.Api
```
brianrob: This is part of the public API so we need to add comments.

Also, can you tell me how you came up with the default values?

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public class PerformanceTestConfig
+    {
+        public PerformanceTestConfig() 
+        {
+            iterations = 30;
+            timeout = 20000;
+        }
+
+        public int iterations;
```
brianrob: These should be public properties with capitalized names per C# coding standard.

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,18 @@
...
+        public int timeout;
```
brianrob: For timeout:

1. Is this a total timeout for all iterations or a per-iteration timeout?  It should be clear based on the name.
2. Rather than using an int, which doesn't tell the user what units are used, I'd recommend a TimeSpan here.  It handles all the complexity for you.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -28,8 +30,10 @@ public XunitPerformanceHarness(string[] args)
             _args = args;
             _disposed = false;
             _outputFiles = new List<string>();
+            _performanceTestConfig = new PerformanceTestConfig();
 
             var options = XunitPerformanceHarnessOptions.Parse(args);
+            _performanceTestConfig.tmpDir = options.TmpDirectory;
```
brianrob: If someone doesn't specify this, what happens?

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
...
+        {
+            pre(_performanceTestConfig, psi);
+            int iterations = _performanceTestConfig.iterations;
+            int timeout = _performanceTestConfig.timeout;
+            psi.RedirectStandardOutput = true;
+
+            // Warmup run.
+            using (var p = new Process()) 
+            {
+                p.StartInfo = psi;
```
brianrob: Looks like this pattern gets used a couple of times here.  Perhaps it would be worthwhile to put it in a helper method and call it.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
...
+                p.Start();
+                if (p.WaitForExit(timeout) == false) 
+                {
+                    if (p != null)
+                    {
+                        p.Kill();
+                    }
+                    Console.WriteLine("Timeouted!");
```
brianrob: We should have a better error message here if something goes wrong.  Also, you should write it to stderr (I think it's Console.Error.WriteLine, but not sure).

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
...
+                }
+            }
+
+            for (int i = 0; i < iterations; i++)
+            {
+                using(var p = new Process())
+                {
+                    p.StartInfo = psi;
+                    p.Start();
+                    if (p.WaitForExit(timeout) == false)
```
brianrob: What happens if the process terminates with a non-zero error code?  Do we continue?  We probably should not, but we do need to make sure that clean-up gets run regardless.

src/xunit.performance.api/PerformanceTestConfig.cs
brianrob: You should have some bounds checking to make sure that someone doesn't specify 0 or negative iteration count.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
...
+                            p.Kill();
+                        }
+                        Console.WriteLine("Timeouted!");
+                        return;
+                    }
+                    mid(_performanceTestConfig);
+                }
+            }
+
+            BenchmarkScenario bs = post(_performanceTestConfig);
```
brianrob: What if bs == null.  You should write an error here to make it clear what the user did wrong.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
...
+            bs.Serialize("out.xml");
+
+            var statisticsFileName = $"JitBench-Statistics";
```
brianrob: This is probably left over from your test case.  There shouldn't be anything JitBench specific here.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -15,6 +15,7 @@ internal sealed class XunitPerformanceHarnessOptions
         public XunitPerformanceHarnessOptions()
         {
             _outputDirectory = Directory.GetCurrentDirectory();
+            _tmpDirectory = Directory.GetCurrentDirectory();
```
brianrob: You should probably make a random directory in the output directory.  This ensures that you can easily delete it and that we don't inadvertently mess things up in a directory that we didn't create.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -44,6 +45,31 @@ public string OutputDirectory
...
+                    throw new Exception("The temporal directory name cannot be null, empty or white space.");
+                }
+
+                if (value.Any(c => Path.GetInvalidPathChars().Contains(c)))
+                {
+                    throw new Exception("Specified temporal directory name contains invalid path characters.");
+                }
+
+                _tmpDirectory = Path.IsPathRooted(value) ? value : Path.GetFullPath(value);
+                if (!Directory.Exists(_tmpDirectory))
```
brianrob: We should make sure that this fails gracefully and it's clear to the user what happened if this fails (e.g. they don't have permissions).

brianrob:
cc: @valenis

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
...
+                        if (p != null)
+                        {
+                            p.Kill();
+                        }
+                        Console.WriteLine("Timeouted!");
+                        return;
+                    }
+                    mid(_performanceTestConfig);
+                }
+            }
```
jorive: Why this is not a single loop with (Iterations + 1) number of iterations? I think it would be cleaner, you can also capture the output, and decide to ignore it when computing statistics and uploading to BenchView.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
             s_profiler(assemblyPath, Configuration.RunId, _outputDirectory, runner, collectOutputFilesCallback);
         }
 
+        public void RunStartupBenchmark(string[] args, ProcessStartInfo psi, PreDel pre, MidDel mid, PostDel post)
+        {
+            pre(_performanceTestConfig, psi);
+            int iterations = _performanceTestConfig.iterations;
+            int timeout = _performanceTestConfig.timeout;
+            psi.RedirectStandardOutput = true;
```
jorive: >psi.RedirectStandardOutput = true; [](start = 12, length = 34)

`psi` is user input parameter. I would not change this without letting the user know, or make it explicit on the function header/comment (it is a public API, and we should be transparent).

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,66 @@ public void RunBenchmarks(string assemblyPath)
             s_profiler(assemblyPath, Configuration.RunId, _outputDirectory, runner, collectOutputFilesCallback);
         }
 
+        public void RunStartupBenchmark(string[] args, ProcessStartInfo psi, PreDel pre, MidDel mid, PostDel post)
+        {
+            pre(_performanceTestConfig, psi);
```
jorive: >pre(_performanceTestConfig, psi); [](start = 12, length = 33)

I think that the Setup and Teardown delegates should be used around the benchmark iterations, and not before and after it. It should be use to keep a clean state between each benchmark iteration.

src/xunit.performance.api/XunitPerformanceHarness.cs
jorive: >BenchmarkScenario bs = post(_performanceTestConfig); [](start = 12, length = 52)

Same as above (it should be between iterations).

src/xunit.performance.api/XunitPerformanceHarness.cs
jorive: >Console.WriteLine [](start = 20, length = 17)

In general, please use the `PerformanceLogger` class instead of Console.WriteLine, as we can control the API output with a verbosity flag in the future.

src/xunit.performance.api/XunitPerformanceHarness.cs
jorive: >PreDel pre, MidDel mid, PostDel post [](start = 77, length = 36)

I would suggest replacing the delegates with Actions<PerformanceTestConfig> (this function can become generic in the future too).

src/xunit.performance.api/Model/AssemblyModel.cs
jorive: HINT: We can create a `TableHeader` class with constants.

---
In reply to: [107046360](https://github.com/Microsoft/xunit-performance/pull/208#discussion_r107046360) [](ancestors = 107046360)

src/xunit.performance.api/Model/AssemblyModel.cs
jorive: Is this field (Namespace) used in the data uploaded to BenchView? If not, then we can remove it.

---
In reply to: [107045986](https://github.com/Microsoft/xunit-performance/pull/208#discussion_r107045986) [](ancestors = 107045986)

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +175,143 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
     }
 
     [Serializable]
+    [XmlRoot("BenchmarkScenario")]
```
guhuro: Seems alright to me

src/xunit.performance.api/PerformanceTestConfig.cs
guhuro: Fixed.

src/xunit.performance.api/XunitPerformanceHarness.cs
DrewScoggins: This seems excessive to me.  This means that all the work that is required to setup the test the first time must be re-run every time.  Why could we just not do any per iteration cleanup work in the PostIteration delegate?

src/xunit.performance.api/Model/AssemblyModel.cs
dpodder: I suggested adding support for something like Namespace to allow the scenarios to be grouped when uploading to BenchView, e.g. "RWC.JitBench.MusicStore" instead of just "MusicStore". It doesn't have to be a blocker for Scenario support, but we should implement it soon so that we can more effectively categorize the data into BenchView.

Now that I look at the implementation, though, I wonder whether we can just leverage the implementation from `[Benchmmark]` style perf tests that use the already-implemented `TestModel` class. In this context, `type` is the namespace, `method` is the scenario name, and `name` is the full type+scenario_name+qualifiers (e.g.: type == "RWC.JitBench", method == "MusicStore", name == "RWC.JitBench.MusicStore"). Qualifiers can be things like input data set or scenario workload, etc.

@jorive Any thoughts?

src/xunit.performance.api/Model/AssemblyModel.cs
dpodder: :+1: FYI, `TestModel` is already defined below. Although, I'm in favor of folding the two together unless there's a good reason to diverge. If we really must keep them separate, perhaps `ScenarioTestModel` might be a more appropriate name?

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
dpodder: Suggestion: you can use `System.IO.Path.GetTempPath()` to get the tmp dir in a platform-agnostic way.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -44,6 +45,31 @@ public string OutputDirectory
             }
         }
 
+        [Option("perf:tmpdir", Required = false, HelpText = "Specifies a writable directory to storage temporal files.")]
```
dpodder: Nit: this should be worded as "...to ***store temporary*** files."

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -44,6 +45,31 @@ public string OutputDirectory
...
+        public string TmpDirectory
+        {
+            get { return _tmpDirectory; }
+
+            set
+            {
+                if (string.IsNullOrWhiteSpace(value))
+                {
+                    throw new Exception("The temporal directory name cannot be null, empty or white space.");
```
dpodder: Nit: temporal -> temporary

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -44,6 +45,31 @@ public string OutputDirectory
...
+                }
+
+                if (value.Any(c => Path.GetInvalidPathChars().Contains(c)))
+                {
+                    throw new Exception("Specified temporal directory name contains invalid path characters.");
```
dpodder: Nit: temporal -> temporary

src/xunit.performance.api/PerformanceTestConfig.cs
guhuro: The former one. Renamed to TimeoutPerIteration and changed to TimeSpan type.

src/xunit.performance.api/XunitPerformanceHarness.cs
dpodder: +1; also, I think we should make a copy of any input parameter we modify so we don't modify the caller's copy.

src/xunit.performance.api/XunitPerformanceHarness.cs
dpodder: I agree that I don't think we want to run setup every time (we should at least have a way to do once-only setup for big tasks, like building something). Either way, we want to make sure to document this well.

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Changed the naming of this variables and functions.
Right now yes, we are considering that if we didn't get an exception and we get a proper return in the last delegate, the run succeeded.

src/xunit.performance.api/Model/AssemblyModel.cs
jorive: It makes sense. Then, I believe we should keep it as it can be used to categorize and group tests.

src/xunit.performance.api/Model/AssemblyModel.cs
jorive: I think that calling it `ScenarioTestModel` makes more sense.

src/xunit.performance.api/XunitPerformanceHarness.cs
dpodder: Spoke with @jorive offline; I was convinced otherwise. :smile: The user is free to do one-time setup in their own C# code that is the harness; therefore we have no reason to support anything other than per-iteration setup. Ergo, I vote that we should move the pre (and post below) to happen per iteration.

jorive: OK, let me explain with more detail on why I think this approach is not such a good idea. This is a performance API and not a performance harness, therefore it should only run performance related tasks (stick to the "single responsibility principle" and have a single functionality).

The current design adds an extra functionality to the API with no real value added (the capability to run non-performance-related tasks). Running tasks such as retrieving the repository and building the Jit bench music store benchmarks are not part of this benchmark scenario. Instead, those tasks should be run inside the performance harness, where the user still have control over it.

I believe that the setup and teardown steps should be used to keep the pre/post benchmark steps on a clean state. For example, setup could be used to keep the pre-benchmark step clean and the teardown to collect data. In short, it should be something needed to ensure the benchmark state is clean in every iteration (users want to run these steps for every iteration).

The structure I envision would look more like the pseudo code below:

```
var nIterations = Iterations + 1; // number of iterations specified by the user plus warmup.
for (int i = 0; i < nIterations; ++i)
{
    // setup
    //   start profiling
    //     run benchmark
    //   stop profiling
    // teardown
}
```

DrewScoggins: OK, this makes sense.  I am fine with using this design, but we will need to make sure that we document that this is how the the API works, and that any one-time setup code needs to not be in that delegate, but in the test code that you write.

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: You're right, fixed.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
guhuro: Added a check

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
guhuro: That's great, thanks.

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Alright. Removed so that it doesn't modify the psi (the user can still set this if needed)

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
guhuro: Done

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
guhuro: Done

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
guhuro: Done

jorive:
@guhuro Could you please update the **README.md** file with information about this new feature?

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Of course, done

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Checked. Nevertheless, we are still relying on the user to provide us with correct information (including returning a valid ScenarioBenchmark object)

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+                    if (p.WaitForExit(timeout) == false) 
+                    {
+                        if (p != null)
+                        {
+                            p.Kill();
+                        }
+                        Console.WriteLine("Timeouted!");
+                        return;
+                    }
+                    postIterationDel(_performanceTestConfig);
```
jorive: This call could be outside the using block.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+                }
+            }
+
+            ScenarioBenchmark scenarioBenchmark = teardownDel(_performanceTestConfig);
+            if (scenarioBenchmark == null) 
+            {
+                Console.Error.WriteLine ("The Teardown Delegate should return a valid instance of ScenarioBenchmark.");
```
jorive: I would throw an exception here.

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    public class PerformanceTestConfig
+    {
+        public PerformanceTestConfig() 
+        {
+            Iterations = 30;
```
jorive: Do we want default values here? Should these be passed in the constructor instead?

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            set
+            {
+                if (value.TotalMilliseconds > 0) 
+                {
+                    _timeoutPerIteration = value;
+                }
+                else 
+                {
+                    Console.Error.WriteLine("Timeout per iteration should be a positive amount of milliseconds.");
```
jorive: I would prefer to throw an exception here.

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            set
+            {
+                if (value > 1) 
+                {
+                    _iterations = value;
+                }
+                else 
+                {
+                    Console.Error.WriteLine("The amount of iterations should be a positive integer greater than 1.");
```
jorive: Exception here too.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+                {
+                    p.StartInfo = processStartInfo;
+                    p.Start();
+                    if (p.WaitForExit(timeout) == false) 
+                    {
+                        if (p != null)
+                        {
+                            p.Kill();
+                        }
+                        Console.WriteLine("Timeouted!");
```
jorive: We should throw an exception here.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+            scenarioBenchmark.Serialize(Configuration.RunId + "-" + scenarioNamespace + ".xml");
+
+            var mdFileName = Configuration.RunId + "-" + scenarioNamespace + "-Statistics.md";
+            var csvFileName = Configuration.RunId + "-" + scenarioNamespace + "-Statistics.csv";
+
+            var dt = scenarioBenchmark.GetStatistics();
+            var mdTable = MarkdownHelper.GenerateMarkdownTable(dt);
+            MarkdownHelper.Write(mdFileName, mdTable);
+            WriteInfoLine($"Markdown file saved to \"{mdFileName}\"");
+            Console.WriteLine(mdTable);
```
jorive: Could you please change this: Console.WriteLine to this: WriteInfoLine

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Yep

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Sure do

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -39,7 +40,46 @@ public string OutputDirectory
                 _outputDirectory = Path.IsPathRooted(value) ? value : Path.GetFullPath(value);
                 if (!Directory.Exists(_outputDirectory))
                 {
-                    Directory.CreateDirectory(_outputDirectory);
+                    try
```
jorive: I would remove the try/catch block here.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -39,7 +40,46 @@ public string OutputDirectory
...
+
+                if (value.Any(c => Path.GetInvalidPathChars().Contains(c)))
+                {
+                    throw new Exception("Specified temporary directory name contains invalid path characters.");
+                }
+
+                _temporaryDirectory = Path.IsPathRooted(value) ? value : Path.GetFullPath(value);
+                if (!Directory.Exists(_temporaryDirectory))
+                {
+                    try
```
jorive: Same as above. I would remove the try/catch block here.

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Done

src/xunit.performance.api/PerformanceTestConfig.cs
guhuro: Added Exception

src/xunit.performance.api/PerformanceTestConfig.cs
guhuro: Done

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Done

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -160,5 +214,10 @@ private void FreeUnManagedResources()
         private readonly List<string> _outputFiles;
         private readonly List<string> _typeNames;
         private bool _disposed;
+        public delegate void SetupDelegate(PerformanceTestConfig config, ProcessStartInfo processStartInfo);
+        public delegate void PostIterationDelegate(PerformanceTestConfig config);
+        public delegate ScenarioBenchmark TeardownDelegate(PerformanceTestConfig config);
```
dpodder: Are these delegates still used? If not, we should remove them.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -39,7 +40,46 @@ public string OutputDirectory
...
+                    throw new Exception("The temporary directory name cannot be null, empty or white space.");
+                }
+
+                if (value.Any(c => Path.GetInvalidPathChars().Contains(c)))
+                {
+                    throw new Exception("Specified temporary directory name contains invalid path characters.");
+                }
+
+                _temporaryDirectory = Path.IsPathRooted(value) ? value : Path.GetFullPath(value);
+                if (!Directory.Exists(_temporaryDirectory))
```
dpodder: You actually don't need to test for this condition: `Directory.CreateDirectory(path)` already does this internally, and returns as a no-op if `path` already exists as a directory. See <https://docs.microsoft.com/en-us/dotnet/core/api/system.io.directory>.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+            for (int i = 0; i < iterations; i++)
+            {
+                preIterationDel(_performanceTestConfig);
+                using(var p = new Process())
+                {
+                    p.StartInfo = processStartInfo;
+                    p.Start();
+                    if (p.WaitForExit(timeout) == false) 
+                    {
+                        if (p != null)
```
jorive: p is already != null, otherwise the statement p.WaitForExit would have thrown a NullReferenceException. The check can be removed.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
jorive: I noticed a small bug too due to copy/paste:
```cshap
                if (!Directory.Exists(_outputDirectory)) // checking for _outputDirectory
                {
                    try 
                    {
                        Directory.CreateDirectory(_temporaryDirectory); // creating _temporaryDirectory
                    } catch (System.Exception) 
                    {
                        Console.Error.WriteLine("Couldn't create Directory " + _temporaryDirectory);
                        throw;
                    }
                }
```

jorive: Could you please replace all properties with `Directory.CreateDirectory` method call without checking `Directory.Exist`  and without the try/catch block?

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -41,6 +41,7 @@ public sealed class AssemblyModel
 
         internal DataTable GetStatistics()
         {
+            //fixme. Column names should be defined as constant methods
```
brianrob: Before merging, this should either be fixed, or an issue filed and the comment removed.

src/xunit.performance.api/Model/AssemblyModel.cs
brianrob: Ideally, this should be as simple as possible.  I don't think people are going to know what to put here.  I like the idea of making it automatic - otherwise, the default value will always be empty string.

My vote would be to automatically set the test name just like we do for microbenchmarks.  I like the grouping mechanism that @dpodder suggests above.

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +176,118 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
+                {
+                    new XmlSerializer(typeof(ScenarioBenchmark))
+                        .Serialize(sw, this, namespaces);
+                }
+            }
+        }
+
+        internal DataTable GetStatistics()
+        {
+            //fixme. Column names should be defined as constant methods
```
brianrob: Same here.  Fix before merge or file an issue and remove the comment.

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -175,6 +176,118 @@ private static IEnumerable<(string testCaseName, string metric, IEnumerable<doub
...
+                }
+            }
+
+            return dt;
+        }
+    }
+
+    [Serializable]
+    [XmlType("Test")]
+    public sealed class ScenarioTestModel
```
brianrob: @dpodder, @jorive is this the API we want to expose for providing results?  I'm not saying we shouldn't, but it looks very specific to XML and this now becomes a public API.  If we want to change this, we're going to break people (or have a compat burden).  Just want to make sure we feel good about this.

src/xunit.performance.api/PerformanceTestConfig.cs
brianrob: Great question.  I'm torn here.  On one hand, I think default values are good because otherwise people have to decide what they should be.  On the other hand, I'm not sure how we pick good defaults.

My suspicion is that we want to have defaults because we can change them if we discover that they're not good.  However, if we force values through the constructor then we can't change it unless we go change everyone's code.  I think that this is less interesting for timeout.  People can make an educated guess on that.  But iterations is much more contentious and unclear.  Maybe we require timeout and have a default value for iterations?

src/xunit.performance.api/PerformanceTestConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+                }
+            }
+        }
+        public int Iterations
+        {
+            get { return _iterations; }
+
+            set
+            {
+                if (value > 1)
```
brianrob: We should force a warm-up iteration on top of this number.  Just making sure - do we do this already?

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -28,8 +30,10 @@ public XunitPerformanceHarness(string[] args)
             _args = args;
             _disposed = false;
             _outputFiles = new List<string>();
+            _performanceTestConfig = new PerformanceTestConfig();
```
brianrob: I think this is just used for scenario tests.  We should name this ScenarioTestConfiguration so that it matches the RunScenario method.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
             s_profiler(assemblyPath, Configuration.RunId, _outputDirectory, runner, collectOutputFilesCallback);
         }
 
+        public void RunScenario(ProcessStartInfo processStartInfo, Action<PerformanceTestConfig> preIterationDel,
```
brianrob: Nit: Rather than using Action, I think it would be good to declare delegates.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+                        return;
+                    }
+                }
+                postIterationDel(_performanceTestConfig);
+            }
+
+            ScenarioBenchmark scenarioBenchmark = teardownDel(_performanceTestConfig);
+            if (scenarioBenchmark == null) 
+            {
+                throw new NullReferenceException("The Teardown Delegate should return a valid instance of ScenarioBenchmark.")
```
brianrob: You should throw an Exception or InvalidOperationException.  NullReferenceException shouldn't be thrown by code explicitly.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+            }
+
+            string scenarioNamespace = scenarioBenchmark.Namespace;
+
+            scenarioBenchmark.Serialize(Configuration.RunId + "-" + scenarioNamespace + ".xml");
+
+            var mdFileName = Configuration.RunId + "-" + scenarioNamespace + "-Statistics.md";
```
brianrob: The RunId is used for microbenchmarks as the full file name.  We should probably be consistent here and not add the scenario namespace to the file name.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -71,6 +75,56 @@ public void RunBenchmarks(string assemblyPath)
...
+            var csvFileName = Configuration.RunId + "-" + scenarioNamespace + "-Statistics.csv";
+
+            var dt = scenarioBenchmark.GetStatistics();
```
brianrob: @jorive, Should all of this serialization code be merged with microbenchmark serialization code?  Are these the same?

src/xunit.performance.api/XunitPerformanceHarness.cs
brianrob: Ah... I just commented above that we should use these instead of Action/Func.  :)

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
```diff
@@ -15,6 +15,7 @@ internal sealed class XunitPerformanceHarnessOptions
         public XunitPerformanceHarnessOptions()
         {
             _outputDirectory = Directory.GetCurrentDirectory();
+            _temporaryDirectory = Path.GetTempPath();
```
brianrob: Just want to make sure that we clean this up when the run completes, right?  It's quite possible for these scenarios to pull down lots of stuff and we could fill up hard drives quickly.

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
guhuro: You're right!
Changed

src/xunit.performance.api/XunitPerformanceHarness.cs
dpodder: @brianrob Out of curiosity, why are delegates preferable to generic `Action` and `Func` lambdas?

src/xunit.performance.api/Model/AssemblyModel.cs
dpodder: I do believe we could do something less constraining, but I'm not sure how much different the format might look at the end, so in the end I wonder whether the value of iterating on the design further is worth too much. Happy to discuss ideas offline though.

src/xunit.performance.api/PerformanceTestConfig.cs
dpodder: @brianrob How do `[Benchmark]` style tests treat iteration counts? Do you have to say `11` if you want `10` iterations, or does it automatically run one more?

src/xunit.performance.api/XunitPerformanceHarnessOptions.cs
dpodder: @brianrob I'm not sure we want to clean when the run completes; but perhaps scorch it when a new run starts? Cleaning it up automatically means throwing out stuff that is valuable when debugging issues/failures (as a scenario writer).

@guhuro either way, if we cleanup/scorch this, we need to make the default a subdirectory of `Path.GetTempPath()` and not the temp path itself--otherwise we will scorch stuff that other programs are using, and that would be Very Bad.

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: For sure

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: Great, I'll change that

src/xunit.performance.api/XunitPerformanceHarness.cs
guhuro: So that doesn't seem to work. WriteInfoLine needs a string as a parameter, and mdTable is a MarkdownLog.Table.

PR #209, opened 2017-03-22
ETW Tracing API update

jorive:
- Updating ETW tracing library version
- Bug fix: Removing Windows PInvoke from Linux.

### DIFF
--- a/src/xunit.performance.api/ETWProfiler.cs
+++ b/src/xunit.performance.api/ETWProfiler.cs
@@ -35,8 +35,7 @@ namespace Microsoft.Xunit.Performance.Api
         /// <returns></returns>
         public static void Record(string assemblyFileName, string runId, string outputDirectory, Action action, Action<string> collectOutputFilesCallback)
         {
-            bool? isElevated = TraceEventSession.IsElevated();
-            if (!isElevated.HasValue || isElevated.Value == false)
+            if (TraceEventSession.IsElevated() != true)
             {
                 const string errMessage = "In order to profile, application is required to run as Administrator.";
                 WriteErrorLine(errMessage);
--- a/src/xunit.performance.api/ETWProfiler.cs
+++ b/src/xunit.performance.api/ETWProfiler.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using static Microsoft.Xunit.Performance.Api.Common;
 using static Microsoft.Xunit.Performance.Api.Native.Windows.Kernel32;
 using static Microsoft.Xunit.Performance.Api.PerformanceLogger;
 
@@ -36,7 +35,8 @@ namespace Microsoft.Xunit.Performance.Api
         /// <returns></returns>
         public static void Record(string assemblyFileName, string runId, string outputDirectory, Action action, Action<string> collectOutputFilesCallback)
         {
-            if (!IsRunningAsAdministrator)
+            bool? isElevated = TraceEventSession.IsElevated();
+            if (!isElevated.HasValue || isElevated.Value == false)
             {
                 const string errMessage = "In order to profile, application is required to run as Administrator.";
                 WriteErrorLine(errMessage);
--- a/tests/standalonesample/standalonesample.csproj
+++ b/tests/standalonesample/standalonesample.csproj
@@ -13,7 +13,7 @@
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.3-alpha-experimental">
       <IncludeAssets>All</IncludeAssets>
     </PackageReference>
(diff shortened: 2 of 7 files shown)


src/xunit.performance.api/ETWProfiler.cs
```diff
@@ -36,7 +35,8 @@ private static SafeTerminateHandler<TraceEventSession> MakeSafeTerminateTraceEve
         /// <returns></returns>
         public static void Record(string assemblyFileName, string runId, string outputDirectory, Action action, Action<string> collectOutputFilesCallback)
         {
-            if (!IsRunningAsAdministrator)
+            bool? isElevated = TraceEventSession.IsElevated();
+            if (!isElevated.HasValue || isElevated.Value == false)
```
dpodder: I'm curious, what does `null` mean in this context? What happens if we're running on a machine where UAC is completely turned off?

tests/standalonesample/standalonesample.csproj
```diff
@@ -1,27 +1,19 @@
...
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.2-alpha-experimental">
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.3-alpha-experimental">
       <IncludeAssets>All</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.NETCore.App" Version="1.0.1" />
-    <PackageReference Include="Microsoft.NET.Sdk" Version="1.0.0-alpha-20161104-2">
-      <PrivateAssets>All</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="xunit.performance.api" Version="1.0.0-*" />
+    <PackageReference Include="xunit.performance.api" Version="1.0.0-alpha-build0049" />
```
dpodder: Are wildcard versions no longer supported? I thought we wanted to make this `*` so that we always picked up the latest version. (Should we make this a `<ProjectReference>` instead to make sure that we grab whatever we built?)

src/xunit.performance.api/ETWProfiler.cs
jorive: This is a call to the Tracing library API and the code looks like this:
```csharp
// TODO FIX NOW make these private 
internal static bool? IsElevated()
{
#if !NOT_WINDOWS 
    Process process = Process.GetCurrentProcess();
    IntPtr tokenHandle = IntPtr.Zero;
#if DOTNET_CORE // TODO FIX use SafeHandles everywhere, but we have to decide how far back we want to support. 
    if (!OpenProcessToken(process.SafeHandle.DangerousGetHandle(), TOKEN_QUERY, out tokenHandle))
        return null;
#else
    if (!OpenProcessToken(process.Handle, TOKEN_QUERY, out tokenHandle))
        return null;
#endif
    int tokenIsElevated = 0;
    int retSize;
    bool success = GetTokenInformation(tokenHandle, TOKEN_INFORMATION_CLASS.TokenElevation, (IntPtr)(&tokenIsElevated), 4, out retSize);
    CloseHandle(tokenHandle);
    if (!success)
        return null;
    GC.KeepAlive(process);      // TODO get on SafeHandles. 
    return tokenIsElevated != 0;
#else
    return true;
#endif // !NOT_WINDOWS
}
```
If the API fails to get information about elevation, then it returns NULL, otherwise it returns TRUE|FALSE. We have not tested on lower than Window 10, but in theory the API should work on Windows 8, and I do not know what happens in lower versions of Windows. Is this something we should support?

brianrob: I don't think we should try to support any lower OS versions than PerfView does.  That would likely just be painful.

tests/standalonesample/standalonesample.csproj
```diff
@@ -1,27 +1,19 @@
...
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.2-alpha-experimental">
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.3-alpha-experimental">
       <IncludeAssets>All</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.NETCore.App" Version="1.0.1" />
-    <PackageReference Include="Microsoft.NET.Sdk" Version="1.0.0-alpha-20161104-2">
-      <PrivateAssets>All</PrivateAssets>
-    </PackageReference>
     <PackageReference Include="xunit.performance.api" Version="1.0.0-*" />
```
brianrob: Since this is the standalone project, we should probably not put 1.0.0-* in the csproj.  This is what everyone is likely to do if they start from our example, and that means that the invariant that people don't get autoupgraded unless they want to is broken.  I know it's a bit of a pain to update this, but perhaps the script that does the updating (BuildForPublication) should do this as well?

PR #214, opened 2017-04-06
Small fixes

jorive:
- Packing xml documentation of the API to enable intellisense information to end users
  (disable warning for missing documentation on public APIs)
- Fixed a bug on the ScenarioConfiguration where we were using `TimeSpam.Milliseconds` instead of `TimeSpam.TotalMilliseconds`
- Removed unused pre-package target from the xUnit Performance API csproj file

### DIFF
--- a/src/xunit.performance.api/ScenarioConfiguration.cs
+++ b/src/xunit.performance.api/ScenarioConfiguration.cs
@@ -13,27 +13,41 @@ namespace Microsoft.Xunit.Performance.Api
         /// Initializes a new instance of the ScenarioConfiguration class.
         /// </summary>
         /// <param name="timeSpan">The amount of time to wait for one iteration process to exit.</param>
-        /// <param name="iterations">Number of times a benchmark scenario process will be executed.</param>
-        /// <param name="validExitCodes">
-        /// A collection of exit codes that indicates success when the benchmark scenario process terminates.
-        /// If this parameter is not specified, then the only valid exit code will 0.
-        /// </param>
-        public ScenarioConfiguration(TimeSpan timeSpan, int iterations = 10, IEnumerable<int> validExitCodes = null)
+        public ScenarioConfiguration(TimeSpan timeSpan)
         {
-            if (timeSpan.TotalMilliseconds <= 0)
+            if (timeSpan.TotalMilliseconds < 1)
                 throw new InvalidOperationException("The time out per iteration must be a positive number.");
-            if (iterations <= 1)
-                throw new InvalidOperationException("The number of iterations must be greater than 1.");
 
-            Iterations = iterations;
+            Iterations = 10;
             TimeoutPerIteration = timeSpan;
-            ValidExitCodes = validExitCodes != null ? new List<int>(validExitCodes) : new List<int> { 0 };
+            ValidExitCodes = new[] { 0 };
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the ScenarioConfiguration class (deep copy).
+        /// </summary>
+        /// <param name="scenarioConfiguration">An instance of the ScenarioConfiguration class</param>
+        public ScenarioConfiguration(ScenarioConfiguration scenarioConfiguration)
+        {
+            Iterations = scenarioConfiguration.Iterations;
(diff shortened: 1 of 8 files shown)


src/xunit.performance.api/ScenarioConfiguration.cs
```diff
@@ -14,7 +14,7 @@ public sealed class ScenarioConfiguration
         /// <param name="iterations">Number of times a benchmark scenario process will be executed.</param>
         public ScenarioConfiguration(TimeSpan timeSpam, int iterations = 10)
```
dpodder: typo: `timeSpam` sound bad. Should probably be `timeSpan`. :smile:

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -92,21 +96,28 @@ public void RunBenchmarks(string assemblyPath)
             if (configuration == null)
                 throw new ArgumentNullException($"{nameof(configuration)} cannot be null.");
 
+            // Make a copy of the ProcessStartInfo to avoid potential bugs due to changes behind the scenes
+            var startInfo = Clone(processStartInfo);
```
brianrob: I don't think you should make this change.  Adding a dependency on JSON for this doesn't feel right, and honestly, if someone wants to make changes to this, that's just abuse of the API.

src/xunit.performance.api/ScenarioConfiguration.cs
```diff
@@ -10,17 +12,22 @@ public sealed class ScenarioConfiguration
...
         /// </summary>
-        /// <param name="timeSpam">The amount of time to wait for one iteration process to exit.</param>
+        /// <param name="timeSpan">The amount of time to wait for one iteration process to exit.</param>
         /// <param name="iterations">Number of times a benchmark scenario process will be executed.</param>
-        public ScenarioConfiguration(TimeSpan timeSpam, int iterations = 10)
+        /// <param name="validExitCodes">
+        /// A collection of exit codes that indicates success when the benchmark scenario process terminates.
+        /// If this parameter is not specified, then the only valid exit code will 0.
+        /// </param>
+        public ScenarioConfiguration(TimeSpan timeSpan, int iterations = 10, IEnumerable<int> validExitCodes = null)
```
brianrob: I think you should separate this out into a mutable and immutable version.  That way, you can have mutable state before the harness method is invoked, and you don't need to make changes to the constructor every time you want to have a new parameter.

src/xunit.performance.api/ScenarioConfiguration.cs
```diff
@@ -13,27 +13,41 @@ public sealed class ScenarioConfiguration
...
+            if (timeSpan.TotalMilliseconds < 1)
                 throw new InvalidOperationException("The time out per iteration must be a positive number.");
-            if (iterations <= 1)
-                throw new InvalidOperationException("The number of iterations must be greater than 1.");
 
-            Iterations = iterations;
+            Iterations = 10;
             TimeoutPerIteration = timeSpan;
-            ValidExitCodes = validExitCodes != null ? new List<int>(validExitCodes) : new List<int> { 0 };
+            ValidExitCodes = new[] { 0 };
```
brianrob: Upon looking at this again, should we call this SuccessExitCodes?

src/xunit.performance.api/ScenarioConfiguration.cs
```diff
@@ -13,27 +13,41 @@ public sealed class ScenarioConfiguration
...
+        /// <param name="scenarioConfiguration">An instance of the ScenarioConfiguration class</param>
+        public ScenarioConfiguration(ScenarioConfiguration scenarioConfiguration)
+        {
+            Iterations = scenarioConfiguration.Iterations;
+            TimeoutPerIteration = scenarioConfiguration.TimeoutPerIteration;
+            ValidExitCodes = scenarioConfiguration.ValidExitCodes;
         }
 
         /// <summary>
         /// Number of times a benchmark scenario process will be executed.
```
brianrob: We might want to describe the warm-up case here.  E.g. if 1 iteration is specified, then no warm-up will occur.  Otherwise, we'll consider the first iteration as warm-up and discard its results.

PR #223, opened 2017-04-28
Adding command line options to override the captured metrics.

jorive:
--perf:collect [metric1[+metric2[+...]]]

  default
    Set by the test author (This is the default behavior if no option is specified. It will also enable ETW to capture some of the Microsoft-Windows-DotNETRuntime tasks).

  stopwatch
    Capture elapsed time using a Stopwatch (It does not require ETW).

  BranchMispredictions|CacheMisses|InstructionRetired
    These are performance metric counters and require ETW.

  gcapi
    It currently enable "Allocation Size on Benchmark Execution Thread" and it is only available through ETW.

Examples
  --perf:collect default
    Collect metrics specified in the test source code by using xUnit Performance API attributes

  --perf:collect BranchMispredictions+CacheMisses+InstructionRetired
    Collects PMC metrics

  --perf:collect stopwatch
    Collects duration

  --perf:collect default+BranchMispredictions+CacheMisses+InstructionRetired+gcapi
    '+' implies union of all specified options


src/xunit.performance.execution/BenchmarkEventSource.cs
```diff
@@ -191,14 +201,24 @@ public unsafe void BenchmarkIterationStop(string RunId, string BenchmarkName, in
                 }
             }
 
-            if (_csvWriter != null)
+            if (s_csvWriter != null)
                 WriteCSV(BenchmarkName, iteration: Iteration);
         }
 
-        public static readonly BenchmarkEventSource Log = new BenchmarkEventSource();
+        private BenchmarkEventSource() : base(true)
```
brianrob: Does this cause EventSource to throw?

src/xunit.performance.metrics/BenchmarkDurationEvaluator.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
brianrob: It looks like you're just splitting these files up.  I'd be interested to understand why, but I don't have a problem with this.

src/xunit.performance.execution/BenchmarkEventSource.cs
jorive: Yes, it does cause EventSource to throw if there is an error on the underlying window code. I can remove it as I instrumented the API to find out why CoreCLR repo stopped reporting events.

brianrob: Interesting... I was thinking that we don't recommend that this be on in production, but I'm tempted to say maybe it should be on because we want to know if there is a failure.  Perhaps you should leave this as is.

src/xunit.performance.metrics/BenchmarkDurationEvaluator.cs
jorive: I was planning to split and re-arrange: attributes vs. metrics vs. evaluators, but it turned out to be a bigger task and I needed this change to go in first.

brianrob: Ok, sounds good.  Go ahead and merge.

src/xunit.performance.execution/BenchmarkEventSource.cs
jorive: In the case of coreclr perf tests, the API did not throw an exception, but `EventSource.IsEnable` always returned false.

src/xunit.performance.metrics/BenchmarkDurationEvaluator.cs
jorive: Should I undo the file split? I'm OK either way.

brianrob: I think if it's easy to undo, then yes, since it's not complete.  Then go ahead and merge.

PR #224, opened 2017-05-02
Updating README for the current release version build0004.

jorive:
- Adding instructions on the metric-collection option `--perf:collect <metric>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,21 +198,53 @@ namespace SampleApiTest
 }
 ```
 
+## Command line options to control the collection of metrics
+
+```
+--perf:collect [metric1[+metric2[+...]]]
+
+    default
+        Set by the test author (This is the default behavior if no option is specified. It will also enable ETW to capture some of the Microsoft-Windows-DotNETRuntime tasks).
+
+    stopwatch
+        Capture elapsed time using a Stopwatch (It does not require ETW).
+
+    BranchMispredictions|CacheMisses|InstructionRetired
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -198,19 +198,50 @@ namespace SampleApiTest
...
+**Cache Misses**                      | Performance Monitor Counter              | Enabled if the collection option `CacheMisses` is specified and the counter is available on the machine <br/>(It requires to run as Administrator)
 **Duration**                          | Benchmark execution time in milliseconds | Always enabled
 **GC Allocations**                    | GC trace event                           | Use the `[MeasureGCAllocations]` attribute in the source code
 **GC Count**                          | GC trace event                           | Use the `[MeasureGCCounts]` attribute in the source code
-**Instructions Retired**              | Performance Monitor Counter              | Use the `[MeasureInstructionsRetired]` attribute in the source code
+**Instructions Retired**              | Performance Monitor Counter              | Enabled if the collection option `InstructionRetired` is specified or the `[MeasureInstructionsRetired]` attribute is used in the source code, and the counter is available on the machine <br/>(It requires to run as Administrator)
```
DrewScoggins: Might want to also caveat here that even if the attribute is specified and we call with just stopwatch we will not collect this metric.

DrewScoggins:
Apart from the one comment LGTM.

README.md
jorive: Right! Thanks :)

Issue #231, opened 2017-05-23
API fails to find tests on Linux when `MeasureInstructionsRetiredAttribute` is specified.


jorive:
The `PerformanceTestMessageVisitor.Visit` method silently throws an exception when the `MeasureInstructionsRetiredAttribute` attribute is found due to an exception thrown by the `TraceEventProfileSources.GetInfo` method (PInvoke exception for a Window dll not found).

jorive:
Fixed with commit https://github.com/Microsoft/xunit-performance/pull/232

PR #232, opened 2017-05-23
Fixing Linux bugs

jorive:
- API was silently failing on Linux when `MeasureInstructionsRetired` was specified (https://github.com/Microsoft/xunit-performance/issues/231).
- Currently, signing on Linux is not possible (https://github.com/dotnet/roslyn/issues/8210). Updated build scripts to partially allow us to build project on Linux.
- Modified build scripts to write commit hash to the nuget packages we create.


AndyAyersMS:
Does this fix Linux to the point where we'll get perf data? Or is there more still to do?

jorive:
@AndyAyersMS Yes, with these changes, we will be able to stopwatch on Linux (no ETW).

PR #236, opened 2017-06-06
Cleaning and unifying csproj files

jorive:
- Part of this work is to enable building the project on Linux.


jorive:
Looks like something changed on the way Jenkins runs the jobs. Now, I get this on the run:
`[ERROR] Cannot run simpleharness test because this is not an administrator window`

jorive:
@dotnet-bot help

PR #239, opened 2017-08-21
Avoid calling StopMeasurement twice.

jorive:
- Remove unnecessary package reference
- Remove global.json file
- Skip attempt to process csv file when there is no file (Benchmarks did not run. For example, they are all facts or disabled)
  (We throw `System.IO.FileNotFoundException` exception when no benchmark ran.)
- Print more Debug information on the ETW profiling.

### DIFF
--- a/BuildForPublication.cmd
+++ b/BuildForPublication.cmd
@@ -5,7 +5,7 @@
 setlocal EnableDelayedExpansion
   set errorlevel=
   set BuildConfiguration=Release
-  set VersionSuffix=beta-build0007
+  set VersionSuffix=beta-build0008
 
   REM Check that git is on path.
   where.exe /Q git.exe || (
--- a/build.cmd
+++ b/build.cmd
@@ -84,12 +84,12 @@ setlocal
 
 :dotnet_pack
 setlocal
+  call :dotnet_build || exit /b 1
+
   echo/
   echo/  ==========
   echo/   Packing %cd%
   echo/  ==========
-  call :dotnet_build || exit /b 1
-
   set MsBuildArgs=
   set "MsBuildArgs=%MsBuildArgs% --no-build"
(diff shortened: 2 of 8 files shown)


brianrob:
Do we have a case where this is an issue?  I would really rather not add additional overhead inside of the benchmark loop.  We already check for duplicate calls here: https://github.com/Microsoft/xunit-performance/blob/master/src/xunit.performance.execution/BenchmarkTestInvoker.cs#L236, so I don't believe this change will actually have an impact (other than avoiding a extra call).

jorive:
I think we should be checking as soon as it happens. We previously Dispose, but as in the small test I added, users might call it twice.
I found this as I was trying to clean and slim down some of the API.

brianrob:
Unfortunately, this means that we now increase our overhead all the time for a case that likely doesn't happen very often.  I would not optimize for this case, but instead assume that Dispose will only be called once.  If it is called twice, then the benchmark is already suspect because it's trying to stop measurement twice.  Arguably, if you call stop twice, we should throw.  I see that in debug an assert is fired.  I'd actually be OK changing that to a throw.

jorive:
In that case, I think I should move the check to the Dispose. What do you think?

brianrob:
On the surface that sounds OK, but then when I look at the code, you're going to have to expose the iteration number so that it can be used in Dispose, since StopMeasurement validates the iteration number.  Then, there's also a Boolean that tracks whether or not start has been called.  This will have to be moved as well.  The more I look at this, the more complexity I see here and the more this doesn't seem to make the implementation leaner, even if it would be nice to have the base class do the check.

AndyAyersMS:
Agree with Brian here, seems like we should just fail nosily if the harness is mis-used.

jorive:
Closing this PR in favor of other changes. I'll address this issue later.

PR #241, opened 2017-09-26
Scenario benchmark upgrade

jorive:
- Bump the verion of the new nuget package
- Generalize the Etw listener, so it can be used in the scenario case too
- Impreved the scenario interface. It provides more options to the scenario user to customize the process that will be launched
- When running an scenario, it will return the Process "Elapsed time"
- When running Etw scenario, it will return the Etw process running time
- When running Etw scenario, data for the specified process, its child processes, and modules will be collected
- Added an Etw simple parser. It collects PMC counters per process, and per module
- Small changes to the AssemblyModel, which was initially not planned to be exposed as public
- Increase the default InstructionRetired.Interval from 100,000 to 1,000,000. This will reduce the overall etl file size and still keep the collected data meanigful
- Created few interfaces such as `PerformanceMonitorCounter` and `ProcessExitInfo` to be exposed through the run scenario
- Small bug fix: `GC.SuppressFinalize` was called before `Dispose(true)`
- Adding a small test for `BenchmarkIterationMeasurement.Dispose`
- Remove unnecessary package reference
- Remove global.json file
- Skip attempt to process csv file when there is no file (Benchmarks did not run. For example, they are all facts or disabled)
  (We throw `System.IO.FileNotFoundException` exception when no benchmark ran.)
- Print more Debug information on the ETW profiling.


jorive:
Pending:
- There is some cleanup I'm doing like removing duplicated code
- Update the README

AndyAyersMS:
Hmm, that is a lot of changes. Anything you want me to focus on?

I looked at bit at the module attribution for PMC events. I am curious where this is going. Are you thinking of handling the Method Load & Unload events from the CLR too?

jorive:
@AndyAyersMS In this PR I'm collecting PMC data from the scenarios via ETW. The plan is to extend it and collect user specified providers data too.

With these changes we will be able to collect PMC data from the TP and benchmarks scenario runs.

src/xunit.performance.api/Profilers/Etw/EtwKernelProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            // FIXME: It is not clear we want these as the default, but when kernel profiling is needed. Instead use the commented flags below?
+            Default = new EtwKernelProvider {
+                Flags = KernelTraceEventParser.Keywords.ImageLoad
+                    | KernelTraceEventParser.Keywords.Process
+                    | KernelTraceEventParser.Keywords.Thread,
+                StackCapture = KernelTraceEventParser.Keywords.ContextSwitch
+                    | KernelTraceEventParser.Keywords.Profile,
+            };
+            //Default = new EtwKernelProvider {
```
brianrob: Nit: Remove commented code.

src/xunit.performance.api/Profilers/Etw/EtwListener.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">The type of the return value of the method that the runner delegate encapsulates.</typeparam>
+    internal sealed class EtwListener<TResult> : ICanListenEvents<TResult>
+    {
+        public EtwListener(
+            EtwSessionData userSessionData,
+            IReadOnlyCollection<EtwUserProvider> userProviders,
+            IReadOnlyCollection<EtwKernelProvider> kernelProviders)
+        {
+            UserSessionData = userSessionData ?? throw new ArgumentNullException($"{nameof(userSessionData)} cannot be null.");
```
brianrob: Nit: parameter to the exception should just be the parameter name.

src/xunit.performance.api/Profilers/Etw/EtwListener.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                kernelSession?.EnableKernelProvider(kernelProvider.Flags, kernelProvider.StackCapture);
+
+                using (var userSession = new EtwSession(UserSessionData))
+                {
+                    UserProviders.ForEach(provider => {
+                        userSession.EnableProvider(provider.Guid, provider.Level, provider.Keywords, provider.Options);
+                    });
+
+                    result = runner();
```
brianrob: You should wrap this in a try/catch or try/finally block to make sure that clean-up happens.

src/xunit.performance.api/Profilers/Etw/EtwManagedMethod.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
+{
+    internal sealed class EtwManagedMethod
```
brianrob: I would recommend not using the Etw prefix on these classes.  You could imagine getting this data from other profilers, and we'd like to be able to do so, especially on non-windows.

src/xunit.performance.api/Profilers/Etw/EtwManagedModule.cs
```diff
@@ -0,0 +1,38 @@
...
+        public long Id { get; set; }
+
+        public string ILPath { get; set; }
+
+        public string ILFileName { get; set; }
+
+        /// <summary>
+        /// TODO: Should PerformanceMonitorCounterData be exposed via a IReadOnlyDictionary?
+        /// </summary>
+        public IDictionary<int, long> PerformanceMonitorCounterData { get; set; }
```
brianrob: What are the int and long in this context?  We should probably make it clearer what these are.

src/xunit.performance.api/Profilers/Etw/EtwManagedModule.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Collection of loaded methods for this module.
+        /// </summary>
+        internal IList<EtwManagedMethod> Methods { get; set; }
+
+        /// <summary>
+        /// Represents the address range where this method was loaded.
```
brianrob: I think you mean "module" instead of "method".

src/xunit.performance.api/Profilers/Etw/EtwModule.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        /// <summary>
+        /// The fully qualified name of the module file.
+        /// </summary>
+        public string FileName { get; set; }
+
+        /// <summary>
+        /// File name and extension of the module.
+        /// </summary>
+        public string ImageName => Path.GetFileName(FileName);
```
brianrob: Any reason to have both the FileName and Image name?  I'd think you'd want to have the full path to the file, and then you can cut that down whenever you need to.

src/xunit.performance.api/Profilers/Etw/EtwManagedModule.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
+{
+    public sealed class EtwManagedModule
```
brianrob: If I remember correctly from our conversation, you were going to put this behind an interface.  If so, you should make these classes internal and just have a public interface.  Also, I think that you probably want ManagedModule to inherit from Module since its set of properties is strictly a superset of Module.

src/xunit.performance.api/Profilers/Etw/EtwPerfInfoPMCSample.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public int ProcessId { get; set; }
+
+        public ulong InstructionPointer { get; set; }
+
+        public int ProfileSource { get; set; }
+
+        public int ThreadId { get; set; }
+
+        public DateTime TimeStamp { get; set; }
```
brianrob: You should check - I think that using a DateTime here may be costly - you might want to convert this on-the-fly from a QPC timestamp if it is ever requested.

src/xunit.performance.api/Profilers/Etw/EtwProcess.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </summary>
+        public int Id { get; set; }
+
+        /// <summary>
+        /// The time that the associated process was started.
+        /// </summary>
+        /// <remarks>
+        /// It will be null if the process started before the ETW session was enabled.
+        /// </remarks>
+        public DateTime? Start { get; set; }
```
brianrob: Recommend being more descriptive with the names - StartDateTime or StartDate and ExitDateTime or ExitDate.

src/xunit.performance.api/Profilers/Etw/EtwProcess.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// TODO: Should PerformanceMonitorCounterData be exposed via a IReadOnlyDictionary?
+        /// </summary>
+        public IDictionary<int, long> PerformanceMonitorCounterData { get; set; }
+
+        /// <summary>
+        /// The modules that have been loaded by the associated process.
+        /// </summary>
+        public IList<EtwModule> Modules { get; set; }
```
brianrob: I think you were going to merge Modules and ManagedModules into one list so that users don't have to iterate over both and handle duplicates.

src/xunit.performance.api/Scenario.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    public sealed class Scenario : IDisposable
+    {
+        public Scenario(ScenarioConfiguration configuration)
```
brianrob: This looks like a new class.  What was used before this (e.g. what does this replace)?

src/xunit.performance.api/ScenarioConfiguration.cs
```diff
@@ -13,28 +14,23 @@ public sealed class ScenarioConfiguration
...
         /// <param name="timeSpan">The amount of time to wait for one iteration process to exit.</param>
-        public ScenarioConfiguration(TimeSpan timeSpan)
+        /// <param name="startInfo">The data with which to start the scenario.</param>
+        public ScenarioConfiguration(TimeSpan timeSpan, ProcessStartInfo startInfo)
         {
             if (timeSpan.TotalMilliseconds < 1)
                 throw new InvalidOperationException("The time out per iteration must be a positive number.");
 
+            Debug.Assert(
+                !(startInfo.UseShellExecute && (startInfo.RedirectStandardError || startInfo.RedirectStandardInput || startInfo.RedirectStandardOutput)),
```
brianrob: Do you need this assert?  I'd expect the Process class to handle this.

src/xunit.performance.api/ScenarioInfo.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.Diagnostics.Tracing.Session;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    public sealed class ScenarioInfo
```
brianrob: This should have a more clear name - the name should clarify when it is provided and what it is for.  "ScenarioExecutionResult" or something like that.  You get one per iteration, right?  So you could imagine also including Iteration in the name so that you can get a collection of them for a given run of a scenario.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -81,77 +83,140 @@ public void RunBenchmarks(string assemblyFileName)
...
 
             for (int i = 0; i < configuration.Iterations; ++i)
             {
-                preIterationDelegate?.Invoke();
+                using (var scenario = new Scenario(configuration))
+                {
+                    ScenarioInfo scenarioInfo;
 
-                // TODO: Start scenario profiling.
+                    configuration.PreIterationDelegate?.Invoke(scenario);
```
brianrob: We should catch and then fail if any custom user code throws when we don't expect it to.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -81,77 +83,140 @@ public void RunBenchmarks(string assemblyFileName)
...
                     {
-                        process.Kill();
-                        throw new TimeoutException("Running benchmark scenario has timed out.");
+                        scenarioInfo = Run(configuration, scenario);
                     }
 
-                    // Check for the exit code.
-                    if (!configuration.SuccessExitCodes.Contains(process.ExitCode))
-                        throw new Exception($"'{processStartInfo.FileName}' exited with an invalid exit code: {process.ExitCode}");
+                    configuration.PostIterationDelegate?.Invoke(scenarioInfo);
```
brianrob: Same here.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -81,77 +83,140 @@ public void RunBenchmarks(string assemblyFileName)
...
+                                return new PerformanceMonitorCounter(m.DisplayName, psi.Name, m.Unit, psi.ID);
+                            });
+
+                        OutputFileCallback?.Invoke(etlFileName);
+                    }
+                    else
                     {
-                        process.Kill();
-                        throw new TimeoutException("Running benchmark scenario has timed out.");
+                        scenarioInfo = Run(configuration, scenario);
```
brianrob: Same here.

src/xunit.performance.metrics/InstructionsRetiredEvaluator.cs
```diff
@@ -30,6 +30,7 @@ internal void Initialize(PerformanceMetricEvaluationContext context, int profile
                     }
                     else
                     {
+                        // FIXME: We should fail here instead.
```
brianrob: Agreed.

src/xunit.performance.metrics/InstructionsRetiredMetricDiscoverer.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Xunit.Performance
 {
     internal partial class InstructionsRetiredMetricDiscoverer : IPerformanceMetricDiscoverer
     {
-        public const int DefaultInterval = 100000; // Instructions per event.
+        public const int DefaultInterval = 1000000; // Instructions per event.
```
brianrob: I'm thinking that we should probably expose this so that it's clear that if results change by orders of magnitude, it might be this.

src/xunit.performance.api/MetricCollectionFactory/PmcPerformanceMetricFactory.cs
```diff
@@ -40,7 +42,7 @@ public PmcPerformanceMetricFactory(string pmcName)
             }
         }
 
-        public IEnumerable<PerformanceMetric> GetMetrics(string assemblyFileName)
+        public IEnumerable<PerformanceMetric> GetMetrics()
```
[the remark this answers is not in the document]
jorive: This was never used, it is internal, and I am rethinking to generalize some of the code (make it more stand alone).

---
In reply to: [141216232](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r141216232) [](ancestors = 141216232)

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -193,7 +193,7 @@ public sealed class ScenarioBenchmark
         [XmlArray("Tests")]
         public List<ScenarioTestModel> Tests { get; set; }
 
-        public ScenarioBenchmark()
```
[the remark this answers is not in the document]
jorive: Users actually need to use the `ScenarioBenchmark(string name)` ctor. instead, to specify the name of the scenario.

---
In reply to: [141216440](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r141216440) [](ancestors = 141216440)

src/xunit.performance.api/PerformanceMonitorCounter/PerformanceMonitorCounter.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public PerformanceMonitorCounter(string displayName, string name, string unit, int id)
+        {
+            DisplayName = displayName;
+            Name = name;
+            Unit = unit;
+            Id = id;
+        }
+
+        public string DisplayName { get; }
```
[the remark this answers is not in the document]
jorive: The differences are:
Name: "InstructionRetired" // This is the keyword to enable the pmc.
Display Name: "Instructions Retired" // This is the name used on the reports.

---
In reply to: [142809700](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r142809700) [](ancestors = 142809700)

src/xunit.performance.api/Profilers/Etw/EtwAddressRange.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
+{
+    internal class EtwAddressRange
```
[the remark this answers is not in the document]
jorive: If the underlying types make sense, then we could use it on non-Windows (I do not know that yet). This is more of an abstraction of the tracing library. Would it make sense to make it non-Etw?

---
In reply to: [142810045](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r142810045) [](ancestors = 142810045)

src/xunit.performance.api/Profilers/Etw/EtwHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    profileSourceIDs.Add(profInfo.ID);
+                    profileSourceIntervals.Add(Math.Min(profInfo.MaxInterval, Math.Max(profInfo.MinInterval, psi.Interval)));
+                }
+            }
+
+            if (profileSourceIDs.Count > 0)
+            {
+                //
+                // FIXME: This function changes the -pmcsources intervals machine wide.
```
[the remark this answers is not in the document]
jorive: This is not covered in this PR. I believe there is already an opened issue for this, but I like keeping the comment on the code too.
Hint: If you use VS Task List, this can show there.

---
In reply to: [142810222](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r142810222) [](ancestors = 142810222)

src/xunit.performance.api/ProcessExitInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    /// <summary>
+    /// Interface that captures the state of the run scenario.
+    /// </summary>
+    public sealed class ProcessExitInfo
+    {
+        /// <summary>
```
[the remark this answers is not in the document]
jorive: This type is an equivalent class to `ProcessStartInfo` where I gather `Process` data before the object is disposed. Therefore, I kept the name unrelated to the `Scenario*` classes. Does it make sense?

---
In reply to: [142809939](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r142809939) [](ancestors = 142809939)

src/xunit.performance.metrics/InstructionsRetiredMetricDiscoverer.cs
jorive: Agree, but more cleanup needs to happen in order the expose this in a clean manner. The definitions of the Pmc types are in two different assemblies. I would prefer to do this in the future.

---
In reply to: [143036788](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143036788) [](ancestors = 143036788)

jorive:
Yes, I will be adding managed and kernel data on next PR :)

---
In reply to: [332351324](https://github.com/Microsoft/xunit-performance/pull/241#issuecomment-332351324) [](ancestors = 332351324)

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -357,11 +357,18 @@ public void WriteXml(XmlWriter writer)
 
     public sealed class MetricModel
     {
-        public string Name { get; set; }
+        // TODO: Should it be internal? 'MetricModel.Name' is only used to generate the 'XmlElement.Name' & 'XmlAttribute.Name'.
```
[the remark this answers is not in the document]
jorive: This is currently exposed and use by the scenario user. I think I should make this change in future PR.

---
In reply to: [141216533](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r141216533) [](ancestors = 141216533)

src/xunit.performance.api/PerformanceMonitorCounter/GenericPerformanceMonitorCounterEvaluator.cs
```diff
@@ -26,6 +26,7 @@ internal void Initialize(PerformanceMetricEvaluationContext context, int profile
                 }
                 else
                 {
+                    // FIXME: We should fail here instead.
```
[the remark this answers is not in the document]
jorive: This has been the behavior from day 1. I'll postpone this change to future PR.

---
In reply to: [142809590](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r142809590) [](ancestors = 142809590)

src/xunit.performance.metrics/InstructionsRetiredEvaluator.cs
jorive: This has been the behavior from day 1. I'll postpone this change to future PR.

---
In reply to: [143036663](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143036663) [](ancestors = 143036663)

src/xunit.performance.api/Profilers/Etw/EtwListener.cs
jorive: In this case/function, the using blocks will cleanup the necessary objects. The only thing pending is the global state when using Pmc, and this happens before hands. I will handle it in a different PR.

---
In reply to: [143033645](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143033645) [](ancestors = 143033645)

src/xunit.performance.api/Profilers/Etw/EtwManagedMethod.cs
jorive: Makes sense, but I am not sure, at this point, if the internal data/types are applicable to other profilers. I've been looking into other ones, and once I get started it will be more clear what direction to take.
I am thinking using interfaces instead. What do you think?

---
In reply to: [143033861](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143033861) [](ancestors = 143033861)

src/xunit.performance.api/Profilers/Etw/EtwManagedModule.cs
jorive: Yes, I'll be pushing these changes in next commit :)

---
In reply to: [143034762](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143034762) [](ancestors = 143034762)

src/xunit.performance.api/Profilers/Etw/EtwManagedModule.cs
jorive: Yes, I need to clean this interfaces.

---
In reply to: [143034051](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143034051) [](ancestors = 143034051)

src/xunit.performance.api/Scenario.cs
jorive: Before we did not have neither scenario nor scenario configuration. We were passing different parameters.

---
In reply to: [143035848](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143035848) [](ancestors = 143035848)

src/xunit.performance.api/ScenarioConfiguration.cs
jorive: I'll remove it. I was using it for debugging/testing my scenarios. It is a left over.

---
In reply to: [143035933](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143035933) [](ancestors = 143035933)

src/xunit.performance.api/MetricCollectionFactory/PmcPerformanceMetricFactory.cs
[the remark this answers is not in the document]
brianrob: Ok, great.  Thanks.

---
In reply to: [143050920](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143050920) [](ancestors = 143050920,141216232)

src/xunit.performance.api/Profilers/Etw/EtwAddressRange.cs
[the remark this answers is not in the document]
brianrob: Per our conversation, I think it's reasonable to keep these as ETW if they are internal and we can make a call.  For public types, we should think harder because we will support non-ETW profilers.

---
In reply to: [143057784](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143057784) [](ancestors = 143057784,142810045)

src/xunit.performance.api/Profilers/Etw/EtwListener.cs
brianrob: Ok, sounds good.

---
In reply to: [143236199](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143236199) [](ancestors = 143236199,143033645)

src/xunit.performance.api/Profilers/Etw/EtwManagedMethod.cs
brianrob: Yup, interfaces seem reasonable.

---
In reply to: [143237043](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143237043) [](ancestors = 143237043,143033861)

src/xunit.performance.api/Profilers/Etw/EtwPerfInfoPMCSample.cs
brianrob: Looks like you may not have a choice here after all...

---
In reply to: [143034936](https://github.com/Microsoft/xunit-performance/pull/241#discussion_r143034936) [](ancestors = 143034936)

brianrob:
Overall, looks good.  I think there are a few more open issues around try/catch of user code, but perhaps they are all handled via the using blocks?

---
In reply to: [332326182](https://github.com/Microsoft/xunit-performance/pull/241#issuecomment-332326182) [](ancestors = 332326182)

jorive:
Yes, I've tried making all resources implement the disposable interface and placing them in using blocks.

---
In reply to: [334887279](https://github.com/Microsoft/xunit-performance/pull/241#issuecomment-334887279) [](ancestors = 334887279,332326182)

PR #242, opened 2017-10-17
Completing the Etw data collection for performance scenarios.

jorive:
- With these changes the API provides Pmc data for native, .NET, anonymous, unknown, and kernel modules (the important file to look at is `SimpleTraceEventParser.cs`)
- Added comments to classes/methods/file header.
- Renamed classes to minimize redundancy (removed the `Etw` prefix from files/classes as this is already part of the `Etw` namespace where the classes are defined).


jorive:
I just want to clarify that renaming the new scenario interface is meant to keep it clean before a new package gets pushed :)

jorive:
By the way, below there is an example output of JitBench:
```
-------------------------------------------------------
Setting up data standard output/error process handlers.
-------------------------------------------------------
[10/16/2017 6:55:13 PM][INF] $ cmd.exe /c START "SIMULATOR_PIPE" /I /B /WAIT dotnet.exe MusicStore.dll
[10/16/2017 6:55:21 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(0).etl"
Server started in 921ms
Request took 1063ms
-------------------------------------------------------
Setting up data standard output/error process handlers.
-------------------------------------------------------
[10/16/2017 6:55:21 PM][INF] $ cmd.exe /c START "SIMULATOR_PIPE" /I /B /WAIT dotnet.exe MusicStore.dll
[10/16/2017 6:55:28 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(1).etl"
Server started in 853ms
Request took 970ms
-------------------------------------------------------
Setting up data standard output/error process handlers.
-------------------------------------------------------
[10/16/2017 6:55:28 PM][INF] $ cmd.exe /c START "SIMULATOR_PIPE" /I /B /WAIT dotnet.exe MusicStore.dll
[10/16/2017 6:55:35 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(2).etl"
Server started in 973ms
Request took 1037ms
-------------------------------------------------------
Setting up data standard output/error process handlers.
-------------------------------------------------------
[10/16/2017 6:55:35 PM][INF] $ cmd.exe /c START "SIMULATOR_PIPE" /I /B /WAIT dotnet.exe MusicStore.dll
[10/16/2017 6:55:41 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(3).etl"
Server started in 813ms
Request took 990ms
-------------------------------------------------------
Setting up data standard output/error process handlers.
-------------------------------------------------------
[10/16/2017 6:55:41 PM][INF] $ cmd.exe /c START "SIMULATOR_PIPE" /I /B /WAIT dotnet.exe MusicStore.dll
[10/16/2017 6:55:49 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(4).etl"
Server started in 1204ms
Request took 1100ms
-------------------------------------------------------
Setting up data standard output/error process handlers.
-------------------------------------------------------
[10/16/2017 6:55:49 PM][INF] $ cmd.exe /c START "SIMULATOR_PIPE" /I /B /WAIT dotnet.exe MusicStore.dll
[10/16/2017 6:55:56 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(5).etl"
Server started in 848ms
Request took 983ms
-------------------------------------------------------
  Post-Processing scenario data.
-------------------------------------------------------
Parsing: S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(0).etl
Parsing: S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(1).etl
Parsing: S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(2).etl
Parsing: S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(3).etl
Parsing: S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(4).etl
Parsing: S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd(5).etl
[10/16/2017 6:56:33 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd.xml"
[10/16/2017 6:56:33 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd.csv"
[10/16/2017 6:56:33 PM][INF] File saved to: "S:\github.com\jorive\xunit-performance\tests\scenario\bin\debug\netcoreapp1.1\Perf-On-cmd.md"
 JitBench                                              | Metric                | Unit  | Iterations |    Average |    STDEV.S |        Min |        Max
:----------------------------------------------------- |:--------------------- |:-----:|:----------:| ----------:| ----------:| ----------:| ----------:
 dotnet.exe                                            | Duration              |  ms   |     5      |   2552.628 |    206.501 |   2374.719 |   2889.530
 dotnet.exe                                            | Branch Mispredictions | count |     5      | 7.459E+007 | 2.077E+006 | 7.289E+007 | 7.804E+007
 dotnet.exe                                            | Cache Misses          | count |     5      | 1.762E+007 | 2.061E+006 | 1.620E+007 | 2.125E+007
 dotnet.exe                                            | Instructions Retired  | count |     5      | 1.230E+010 | 4.385E+007 | 1.225E+010 | 1.234E+010
 dotnet.exe!Anonymously Hosted DynamicMethods Assembly | Branch Mispredictions | count |     5      |  36864.000 |  12624.720 |  24576.000 |  57344.000
 dotnet.exe!Anonymously Hosted DynamicMethods Assembly | Cache Misses          | count |     5      |  47513.600 |  20191.244 |  28672.000 |  81920.000
 dotnet.exe!Anonymously Hosted DynamicMethods Assembly | Instructions Retired  | count |     4      | 1.750E+006 | 9.574E+005 | 1.000E+006 | 3.000E+006
 dotnet.exe!clrjit.dll                                 | Branch Mispredictions | count |     5      | 2.392E+007 | 2.805E+005 | 2.364E+007 | 2.428E+007
 dotnet.exe!clrjit.dll                                 | Cache Misses          | count |     5      | 2.256E+006 | 6.948E+005 | 1.774E+006 | 3.445E+006
 dotnet.exe!clrjit.dll                                 | Instructions Retired  | count |     5      | 1.661E+009 | 2.263E+007 | 1.637E+009 | 1.687E+009
 dotnet.exe!coreclr.dll                                | Branch Mispredictions | count |     5      | 1.725E+007 | 3.717E+005 | 1.697E+007 | 1.789E+007
 dotnet.exe!coreclr.dll                                | Cache Misses          | count |     5      | 5.047E+006 | 8.625E+005 | 4.411E+006 | 6.545E+006
 dotnet.exe!coreclr.dll                                | Instructions Retired  | count |     5      | 2.025E+009 | 2.875E+007 | 1.977E+009 | 2.054E+009
 dotnet.exe!dotnet.exe                                 | Cache Misses          | count |     1      |   4096.000 |        NaN |   4096.000 |   4096.000
 dotnet.exe!dotnet.exe                                 | Instructions Retired  | count |     1      | 1.000E+006 |        NaN | 1.000E+006 | 1.000E+006
 dotnet.exe!MusicStore.dll                             | Branch Mispredictions | count |     3      |   4096.000 |      0.000 |   4096.000 |   4096.000
 dotnet.exe!MusicStore.dll                             | Cache Misses          | count |     2      |  18432.000 |   2896.309 |  16384.000 |  20480.000
 dotnet.exe!ntoskrnl.exe                               | Branch Mispredictions | count |     5      | 2.115E+007 | 1.196E+006 | 2.014E+007 | 2.312E+007
 dotnet.exe!ntoskrnl.exe                               | Cache Misses          | count |     5      | 4.408E+006 | 4.426E+005 | 3.977E+006 | 5.120E+006
 dotnet.exe!ntoskrnl.exe                               | Instructions Retired  | count |     5      | 7.506E+009 | 4.828E+007 | 7.463E+009 | 7.564E+009
 dotnet.exe!System.Private.CoreLib.dll                 | Branch Mispredictions | count |     5      | 1.864E+006 |  57955.143 | 1.790E+006 | 1.950E+006
 dotnet.exe!System.Private.CoreLib.dll                 | Cache Misses          | count |     5      | 1.027E+006 | 1.090E+005 | 8.643E+005 | 1.126E+006
 dotnet.exe!System.Private.CoreLib.dll                 | Instructions Retired  | count |     5      | 1.680E+008 | 6.964E+006 | 1.570E+008 | 1.750E+008
 dotnet.exe!Unknown                                    | Branch Mispredictions | count |     5      | 1.317E+006 |  38706.658 | 1.249E+006 | 1.343E+006
 dotnet.exe!Unknown                                    | Cache Misses          | count |     5      | 7.045E+005 |  52294.227 | 6.267E+005 | 7.537E+005
 dotnet.exe!Unknown                                    | Instructions Retired  | count |     5      | 9.280E+007 | 3.768E+006 | 8.700E+007 | 9.600E+007
 MusicStore                                            | Duration              |  ms   |     5      |   2580.642 |    206.104 |   2401.352 |   2916.691
 MusicStore/First Request                              | Duration              |  ms   |     5      |   1016.000 |     53.334 |    970.000 |   1100.000
 MusicStore/Startup                                    | Duration              |  ms   |     5      |    938.200 |    160.414 |    813.000 |   1204.000
```

src/xunit.performance.api/ISpan.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+
+namespace Microsoft.Xunit.Performance.Api
+{
+    /// <summary>
+    /// Defines an interval with a mechanism for testing if a value falls in the range.
+    /// </summary>
+    /// <typeparam name="T">The type span space.</typeparam>
+    internal interface ISpan<T>
```
AndyAyersMS: You seem to refer to this as an "interval" in comments, so why not call it `IInterval<T>`?

src/xunit.performance.api/Profilers/Etw/AddressSpace.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            Start = address;
+            Size = size;
+            End = Start + (uint)Size;
+        }
+
+        /// <summary>
+        /// The size of the address space.
+        /// </summary>
+        public int Size { get; }
```
AndyAyersMS: `Size` will always be >= 0, so does `uint` or `ulong` make sense here?

src/xunit.performance.api/Profilers/Etw/AddressSpace.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// Represents an address space as given by the Microsoft.Diagnostics.Tracing.Parsers
+    /// </summary>
+    internal sealed class AddressSpace : ISpan<ulong>
+    {
+        /// <summary>
+        /// Initializes a new instance of the AddressSpace class.
+        /// </summary>
+        /// <param name="address">Interval's start address.</param>
+        /// <param name="size">Interval's size.</param>
+        public AddressSpace(ulong address, int size)
```
AndyAyersMS: Same note as below about size being unsigned.

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+            };
+
+            using (var source = new ETWTraceEventSource(scenarioInfo.EventLogFileName))
+            {
+                if (source.EventsLost > 0)
+                    throw new Exception($"Events lost in trace '{scenarioInfo.EventLogFileName}'");
+
+                ////////////////////////////////////////////////////////////////
+                // Process data
+                var parser = new KernelTraceEventParser(source);
```
AndyAyersMS: I don't understand the asymmetry between start (which looks for parent) and stop (which doesn't). Can you explain?

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+
+                ////////////////////////////////////////////////////////////////
+                // "ntoskrnl.exe" data
+                Func<ImageLoadTraceData, bool> IsNtoskrnlModule = (ImageLoadTraceData obj) => {
+                    return obj.ProcessID == 0 && obj.FileName.EndsWith("ntoskrnl.exe");
+                };
+                parser.ImageDCStart += (ImageLoadTraceData obj) => {
+                    if (IsNtoskrnlModule(obj))
+                    {
+                        if (tmpNtoskrnlModule == null)
```
AndyAyersMS: Calling this a `tmp` kernel module seems odd, since it exists for the duration of the trace. Maybe call it "existing" or "default" or something?

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+                            tmpNtoskrnlModule = new Module(obj.FileName, obj.ImageChecksum);
+
+                        // Assuming nothing else has changed, and keeping the list of already measured Pmc.
```
AndyAyersMS: I'm not sure what this comment means.

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+                                    return;
+                                }
+
+                                var methods = module.Methods
+                                    .Where(m => {
+                                        return m.LifeSpan.IsInInterval(pmc.TimeStamp)
+                                            && m.AddressSpace.IsInInterval(pmc.InstructionPointer);
+                                    })
+                                    .Select(m => m);
+                                if (methods.Count() != 0)
```
AndyAyersMS: There can potentially be hundreds of modules, thousands of methods in a module, and thousands of PMC events, so I worry about efficiency here. Do you have any measurements of how long this processing takes?

By doing this resolution in-stream instead of deferring it you don't need to worry about the time span aspect. And since method ranges generally don't overlap (eg we expect count to be either 0 or 1, and at any rate we don't distinguish between 1 and > 1), when you do this in-stream the whole things is amenable to fast lookup, say by keeping method ranges sorted by starting address.

src/xunit.performance.api/Profilers/Etw/AddressSpace.cs
jorive: I started with the same interface provided by the `TraceEvent` library. I can change it now.

jorive: I started with the same interface provided by the `TraceEvent` library. I can change it now.

---
In reply to: [145194543](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r145194543) [](ancestors = 145194543)

src/xunit.performance.api/Profilers/Etw/AddressSpace.cs
jorive: Same as below

---
In reply to: [145194737](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r145194737) [](ancestors = 145194737)

src/xunit.performance.api/ISpan.cs
jorive: I use span and interval in different places.

---
In reply to: [145194293](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r145194293) [](ancestors = 145194293)

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
jorive: No, I should move this into the `Func` object. This is a left over from cleanup.

---
In reply to: [145197586](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r145197586) [](ancestors = 145197586)

src/xunit.performance.api/Profilers/Etw/AddressSpace.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        /// <param name="lhs">The first AddressSpace to compare, or null.</param>
+        /// <param name="rhs">The second AddressSpace to compare, or null.</param>
+        /// <returns>True if its two operands refer to the same object or if the values of its operands are equal; otherwise, false.</returns>
+        public static bool operator ==(AddressSpace lhs, AddressSpace rhs)
+        {
+            if (ReferenceEquals(lhs, rhs))
+                return true;
+
+            if ((object)lhs == null || (object)rhs == null)
```
brianrob: Any reason not to call AddressSpace.Equals here?  Then you don't have to implement the equality comparison twice (especially if new conditions get added over time).

src/xunit.performance.api/Profilers/Etw/LifeSpan.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <returns>True if its two operands refer to the same object or if the values of its operands are equal; otherwise, false.</returns>
+        public static bool operator ==(LifeSpan lhs, LifeSpan rhs)
+        {
+            if (ReferenceEquals(lhs, rhs))
+                return true;
+
+            if ((object)lhs == null || (object)rhs == null)
+                return false;
+
+            return lhs.Start == rhs.Start && lhs.End == rhs.End;
```
brianrob: Same question as the other interval object - should this just call LifeSpan.Equals?

src/xunit.performance.api/Profilers/Etw/LifeSpan.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// <summary>
+    /// Defines the length of time for which an object lives.
+    /// </summary>
+    public sealed class LifeSpan : IInterval<DateTime>
+    {
+        /// <summary>
+        /// Gets the time associated with the object lifetime start.
+        /// </summary>
+        public DateTime Start { get; set; } = DateTime.MinValue;
```
brianrob: Is it ever reasonable to create one of these objects without specifying both the start and the end?  If not, it would be good to make the set operations private and force the creator to specify them when calling the constructor.

src/xunit.performance.api/Profilers/Etw/Module.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Module's checksum.
+        /// </summary>
+        public int Checksum { get; }
+
+        /// <summary>
+        /// TODO: Should PerformanceMonitorCounterData be exposed via a IReadOnlyDictionary?
+        /// </summary>
+        public IDictionary<PerformanceMonitorCounter, long> PerformanceMonitorCounterData { get; private set; }
```
brianrob: I think that certainly by the time this data makes it back to the consumer it should be read-only.

src/xunit.performance.api/Profilers/Etw/Module.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        internal LifeSpan LifeSpan { get; }
+
+        /// <summary>
+        /// Creates a new object that is a deep copy of the current instance.
+        /// </summary>
+        /// <returns>A new object that is a deep copy of this instance.</returns>
+        internal Module Copy()
+        {
+            var newModule = new Module(FullName, Checksum) {
```
brianrob: Out of curiosity, when do you need to do a deep copy of this object?

src/xunit.performance.api/Profilers/Etw/PerformanceMonitorCounter.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            if (ReferenceEquals(lhs, rhs))
+                return true;
+
+            if ((object)lhs == null || (object)rhs == null)
+                return false;
+
+            return lhs.Id == rhs.Id
+                && lhs.DisplayName == rhs.DisplayName
+                && lhs.Name == rhs.Name
```
brianrob: Same question here.

src/xunit.performance.api/Profilers/Etw/PmcRollover.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
+{
+    /// <summary>
+    /// Defines a performance monitor sample.
+    /// </summary>
+    internal sealed class PmcRollover
```
brianrob: Nit: I'd recommend calling this a PmcSample instead of a Rollover.  It took me a bit of time to realize that (I think) you mean that this object represents a PMC event that is fired once the measured metric threshold is reached.

src/xunit.performance.api/Profilers/Etw/Process.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Life span of this process (From the time that the associated process was started until the time that the associated process exited).
+        /// </summary>
+        public LifeSpan LifeSpan { get; }
+
+        /// <summary>
+        /// TODO: Should PerformanceMonitorCounterData be exposed via a IReadOnlyDictionary?
+        /// </summary>
+        public IDictionary<PerformanceMonitorCounter, long> PerformanceMonitorCounterData { get; }
```
brianrob: I think that this should be read-only by the time it gets back to the consumer.

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+                    if (IsOurProcess(obj))
+                    {
+                        var process = new Process(obj.ImageFileName, obj.ProcessID, obj.ParentID);
+                        process.LifeSpan.Start = obj.TimeStamp;
+                        processes.Add(process);
+                    }
+                };
+                parser.ProcessStop += (ProcessTraceData obj) => {
+                    if (IsOurProcess(obj))
+                        processes.Single(process => process.Id == obj.ProcessID).LifeSpan.End = obj.TimeStamp;
```
brianrob: What will happen here if there are missed events and the process start was never processed?  I suspect it will throw a NRE in which case it might be good to check for this and provide a better error message.

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+                };
+                parser.ImageUnload += (ImageLoadTraceData obj) => {
+                    var process = processes.SingleOrDefault(p => p.Id == obj.ProcessID);
+                    if (process == null)
+                        return;
+
+                    // Check if the unloaded module is on the list.
+                    // The module must be loaded, in the same address space, and same file name.
+                    var module = process.Modules
+                        .SingleOrDefault(m => obj.TimeStamp > m.LifeSpan.End && obj.Equal(m));
```
brianrob: I don't think that this condition can ever evaluate to true and so SingleOrDefault will never return you a valid module.  m.LifeSpan.End starts as DateTime.MaxValue which is always greater than obj.TimeStamp.

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+                ////////////////////////////////////////////////////////////////
+                // .NET modules
+                parser.Source.Clr.LoaderModuleLoad += (ModuleLoadUnloadTraceData obj) => {
+                    var process = processes.SingleOrDefault(p => p.Id == obj.ProcessID);
+                    if (process == null)
+                        return;
+
+                    // FIXME: Is this condition correct? Will a module be always loaded, on ImageLoad, before getting here?
+                    var module = process.Modules
+                        .SingleOrDefault(m => m.FullName == obj.ModuleILPath);
```
brianrob: I think that this is fine on core but would be broken on desktop.  The ModuleILPath and the ModuleNativePath are equivalent on core, but they are different on desktop.  I believe that you will only ever actually use one of them.  The native image contains the IL so even if a method isn't pre-compiled, the IL can be pulled from the native image.  You may wish to see if the ModuleNativePath is specified and use that first before the ModuleILPath.  I believe that ModuleNativePath is only specified if the we were able to bind to it.

src/xunit.performance.api/Profilers/Etw/DotNetModule.cs
```diff
@@ -8,29 +8,29 @@ namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
...
-        /// Module Id
+        /// Module Id.
         /// </summary>
         public long Id { get; }
 
         /// <summary>
         /// Collection of loaded methods for this module.
         /// </summary>
-        internal IList<EtwDotNetMethod> Methods { get; }
+        internal IList<DotNetMethod> Methods { get; }
```
brianrob: Do we ever want to expose per-method counts?  I'm wondering if because we're sampling if this will be too coarse to want to use in a benchmarking system.

src/xunit.performance.api/ScenarioExecutionResult.cs
```diff
@@ -1,23 +1,39 @@
...
+        /// Binary .etl file name where the ETW traces were logged.
+        /// </summary>
         public string EventLogFileName { get; set; }
 
         /// <summary>
         /// This is the list of Pmc that were captured by the ETW event listener.
-        /// TODO: We should reconsider this type, as it is internal data and it does not contain the display name.
         /// </summary>
-        public IEnumerable<PerformanceMonitorCounter> PerformanceMonitorCounters { get; set; }
+        public ISet<PerformanceMonitorCounter> PerformanceMonitorCounters { get; set; }
```
brianrob: Based on our conversation, I'm trying to rationalize the model for returned results.  I think it is OK.  You basically have a set of processes that each have a set of modules.  Both of these objects contain the PMC data that was aggregated for each of them.  For example, each module has PMC data that showed up for that module.  Each process has PMC data for all of the modules that accrued PMC data in that process.  Is that right?

src/xunit.performance.api/Profilers/Etw/LifeSpan.cs
jorive: I added one method before the other. I'll clean it up :)

---
In reply to: [146050754](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146050754) [](ancestors = 146050754)

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
```diff
@@ -0,0 +1,378 @@
...
+                            defaultNtoskrnlModule = new Module(obj.FileName, obj.ImageChecksum);
+
+                        defaultNtoskrnlModule.AddressSpace = new AddressSpace(obj.ImageBase, (uint)obj.ImageSize);
+                        defaultNtoskrnlModule.LifeSpan.Start = obj.TimeStamp;
+                        defaultNtoskrnlModule.LifeSpan.End = DateTime.MaxValue;
+                    }
+                };
+                parser.ImageDCStop += (ImageLoadTraceData obj) => {
+                    if (IsNtoskrnlModule(obj)
+                        && defaultNtoskrnlModule != null
```
brianrob: You probably don't need to process ImageDCStop events.  DCStop events are fired at the end of the process or trace so you are guaranteed that the module is loaded until the end of the process.

src/xunit.performance.api/Profilers/Etw/Module.cs
jorive: Currently it is only used to clone the default Kernel module. I think I'll move it out of here.

---
In reply to: [146051928](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146051928) [](ancestors = 146051928)

src/xunit.performance.api/Profilers/Etw/PerformanceMonitorCounter.cs
jorive: Same as above. I'll clean it up.

---
In reply to: [146052047](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146052047) [](ancestors = 146052047)

src/xunit.performance.api/Profilers/Etw/PmcRollover.cs
jorive: Yes, that is exactly what it is.

---
In reply to: [146052342](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146052342) [](ancestors = 146052342)

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
jorive: I throw an exception earlier if there are lost events. In addition, we only get here if the process has been added to the list.

---
In reply to: [146053767](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146053767) [](ancestors = 146053767)

src/xunit.performance.api/Profilers/Etw/DotNetModule.cs
jorive: Yes, I could expose it. I've been very conservative on the things I make public, because I'm working around the needs of the CoreClr performance scenarios only, but still designing a reusable interface that makes sense for most cases.

---
In reply to: [146059550](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146059550) [](ancestors = 146059550)

jorive: I was thinking that we could also collect "number of jitted methods" as a metric (I already have that information).

---
In reply to: [146065630](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146065630) [](ancestors = 146065630,146059550)

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
jorive: You are right, this is wrong. I'll fix it.

---
In reply to: [146055343](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146055343) [](ancestors = 146055343)

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
jorive: Yes, I did noticed that `obj.ModuleNativePath` is an empty string most of the times.

---
In reply to: [146057410](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146057410) [](ancestors = 146057410)

jorive: Do you mean something like this:
`var modulePath = string.IsNullOrEmpty(obj.ModuleNativePath) ? obj.ModuleILPath : obj.ModuleNativePath;`

---
In reply to: [146083680](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146083680) [](ancestors = 146083680,146057410)

src/xunit.performance.api/ScenarioExecutionResult.cs
jorive: Yes, that is correct.

---
In reply to: [146059975](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r146059975) [](ancestors = 146059975)

src/xunit.performance.api/Profilers/Etw/SimpleTraceEventParser.cs
jorive: Currently, it takes ~7 sec to process a 280MB etl file (MusicStore.dll on CoreClr).

---
In reply to: [145202704](https://github.com/Microsoft/xunit-performance/pull/242#discussion_r145202704) [](ancestors = 145202704)

AndyAyersMS: That seems reasonably fast. W can defer working on any improvements here until we see examples that are too slow.

AndyAyersMS:
I have a bunch of other potential asks but I'd like to see this propagate over to CoreCLR and work with it a bit before I come back with particulars.

Changes LGTM but Brian should have the final say.

PR #243, opened 2017-10-27
Add more capabilities to scenario tests

dsplaisted:
This makes some changes to scenario testing:

- Multiple tests can be run under a single `ScenarioBenchmark`
- Statistics output can be deferred in order to aggregate the results from multiple scenarios
- When creating tables for the results, it is possible to have a column for the scenario name

An example of code that uses the updated capabilities is here: https://github.com/dsplaisted/sdk/blob/17903b0e263aa0835d333f465972889b82670c21/test/Microsoft.NET.Perf.Tests/PerfTest.cs

It creates results like this:

Scenario Name             | Test Name          | Metric         | Unit | Iterations |    Average |  STDEV.S |        Min |        Max
:------------------------- |:------------------ |:-------------- |:----:|:----------:| ----------:| --------:| ----------:| ----------:
.NET Core 2 Console App   | Build              | Execution Time |  ms  |     9      |   2288.169 |   94.770 |   2155.042 |   2404.362
.NET Core 2 Console App   | Build (no changes) | Execution Time |  ms  |     9      |   1220.919 |   33.254 |   1165.615 |   1263.797
.NET Core 2 Console App   | Restore (No-op)    | Execution Time |  ms  |     9      |    797.782 |   16.235 |    765.929 |    817.410
.NET Standard 2.0 Library | Build              | Execution Time |  ms  |     9      |   1950.159 |  118.927 |   1824.096 |   2096.167
.NET Standard 2.0 Library | Build (no changes) | Execution Time |  ms  |     9      |    952.692 |   25.762 |    911.742 |    992.312
.NET Standard 2.0 Library | Restore (No-op)    | Execution Time |  ms  |     9      |    836.623 |   42.447 |    788.992 |    919.064
ASP.NET Core MVC app      | Build              | Execution Time |  ms  |     9      |   3702.768 |   64.516 |   3576.693 |   3827.518
ASP.NET Core MVC app      | Build (no changes) | Execution Time |  ms  |     9      |   2415.289 |   71.831 |   2330.799 |   2529.900
ASP.NET Core MVC app      | Restore (No-op)    | Execution Time |  ms  |     9      |   1123.686 |   24.170 |   1087.291 |   1154.485
LargeP2POldCsproj         | Build              | Execution Time |  ms  |     9      |  58004.077 | 2665.748 |  54163.415 |  62418.122
LargeP2POldCsproj         | Build (no changes) | Execution Time |  ms  |     9      |  10702.538 |  206.794 |  10459.133 |  10995.135
LargeP2POldCsproj         | Restore (No-op)    | Execution Time |  ms  |     9      |   5949.519 |  188.798 |   5757.808 |   6278.338
Roslyn                    | Build              | Execution Time |  ms  |     9      | 165812.729 | 9708.248 | 148006.225 | 177167.230
Roslyn                    | Build (no changes) | Execution Time |  ms  |     9      |  18321.198 | 1044.836 |  16394.047 |  19167.956
SmallP2PNewCsproj         | Build              | Execution Time |  ms  |     9      |   9009.427 |  921.782 |   7347.273 |   9809.446
SmallP2PNewCsproj         | Build (no changes) | Execution Time |  ms  |     9      |   5039.876 |  413.148 |   4309.214 |   5543.195
SmallP2PNewCsproj         | Restore (No-op)    | Execution Time |  ms  |     9      |   1550.481 |   46.974 |   1467.509 |   1610.993
SmallP2POldCsproj         | Build              | Execution Time |  ms  |     9      |   4549.422 |  122.519 |   4413.579 |   4796.330
SmallP2POldCsproj         | Build (no changes) | Execution Time |  ms  |     9      |   1401.622 |   30.569 |   1354.474 |   1459.674
SmallP2POldCsproj         | Restore (No-op)    | Execution Time |  ms  |     9      |    688.694 |   48.489 |    626.152 |    774.784

### DIFF
--- a/src/xunit.performance.api/Model/AssemblyModel.cs
+++ b/src/xunit.performance.api/Model/AssemblyModel.cs
@@ -324,7 +324,7 @@ namespace Microsoft.Xunit.Performance.Api
         }
     }
 
-    public sealed class ScenarioTestResultRow
+    internal sealed class ScenarioTestResultRow
     {
         public string ScenarioName { get; set; }
         public string TestName { get; set; }
--- a/src/xunit.performance.api/Model/AssemblyModel.cs
+++ b/src/xunit.performance.api/Model/AssemblyModel.cs
@@ -222,17 +222,66 @@ namespace Microsoft.Xunit.Performance.Api
             }
         }
 
-        internal DataTable GetStatistics()
+        internal static DataTable GetEmptyTable(string scenarioName = null)
         {
             var dt = new DataTable();
-            var col0_testName = dt.AddColumn(Name);
-            var col1_metric = dt.AddColumn(TableHeader.Metric);
-            var col2_unit = dt.AddColumn(TableHeader.Unit);
-            var col3_iterations = dt.AddColumn(TableHeader.Iterations);
-            var col4_average = dt.AddColumn(TableHeader.Average);
-            var col5_stdevs = dt.AddColumn(TableHeader.StandardDeviation);
-            var col6_min = dt.AddColumn(TableHeader.Minimum);
-            var col7_max = dt.AddColumn(TableHeader.Maximum);
+            if (scenarioName == null)
+            {
+                dt.AddColumn(TableHeader.ScenarioName);
+                dt.AddColumn(TableHeader.TestName);
+            }
+            else
+            {
+                dt.AddColumn(scenarioName);
+            }
+            dt.AddColumn(TableHeader.Metric);
+            dt.AddColumn(TableHeader.Unit);
+            dt.AddColumn(TableHeader.Iterations);
+            dt.AddColumn(TableHeader.Average);
+            dt.AddColumn(TableHeader.StandardDeviation);
+            dt.AddColumn(TableHeader.Minimum);
+            dt.AddColumn(TableHeader.Maximum);
+
+            return dt;
+        }
+
+        internal DataTable GetStatisticsTable(bool includeScenarioNameColumn = false)
+        {
+            var dt = GetEmptyTable(includeScenarioNameColumn ? null : Name);
+
+            AddRowsToTable(dt, GetStatistics(), includeScenarioNameColumn);
+
+            return dt;
+        }
+
+        internal void AddRowsToTable(DataTable dt, IEnumerable<ScenarioTestResultRow> rows, bool includeScenarioNameColumn = false)
+        {
+            foreach (var row in rows)
+            {
+                var newRow = dt.AppendRow();
+                if (includeScenarioNameColumn)
+                {
+                    newRow[dt.ColumnNames[TableHeader.ScenarioName]] = row.ScenarioName;
+                    newRow[dt.ColumnNames[TableHeader.TestName]] = row.TestName;
+                }
+                else
+                {
+                    newRow[dt.ColumnNames[row.ScenarioName]] = row.TestName;
+                }
+                newRow[dt.ColumnNames[TableHeader.Metric]] = row.MetricName;
+                newRow[dt.ColumnNames[TableHeader.Unit]] = row.MetricUnit;
+
+                newRow[dt.ColumnNames[TableHeader.Iterations]] = row.Iterations.ToString();
+                newRow[dt.ColumnNames[TableHeader.Average]] = row.Average.ToString();
+                newRow[dt.ColumnNames[TableHeader.StandardDeviation]] = row.StandardDeviation.ToString();
+                newRow[dt.ColumnNames[TableHeader.Minimum]] = row.Minimum.ToString();
+                newRow[dt.ColumnNames[TableHeader.Maximum]] = row.Maximum.ToString();
+            }
+        }
+
+        internal List<ScenarioTestResultRow> GetStatistics()
+        {
+            List<ScenarioTestResultRow> ret = new List<ScenarioTestResultRow>();
 
             foreach (var test in Tests)
             {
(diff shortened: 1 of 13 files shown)


jorive:
Could you please add the standard C# docstring to the newly added public properties on the `ScenarioTestConfiguration` and `ScenarioExecutionResult` classes? We haven't been good at it in the past, but I've been trying to keep up with it :)

jorive:
If you have a simple (relatively short) scenario example, could you please add it as a test?

src/xunit.performance.api/ScenarioTestConfiguration.cs
```diff
@@ -78,13 +78,20 @@ public IEnumerable<int> SuccessExitCodes
...
         /// <summary>
         /// The action that will be executed after every benchmark scenario execution.
         /// </summary>
         public Action<ScenarioExecutionResult> PostIterationDelegate { get; set; }
 
+        public string TestName { get; set; }
+
+        public ScenarioBenchmark Scenario { get; set; }
+
+        public bool SaveResults { get; set; }
```
jorive: >SaveResults [](start = 20, length = 11)

I think default should be On

src/xunit.performance.api/Model/AssemblyModel.cs
```diff
@@ -254,24 +303,40 @@ internal DataTable GetStatistics()
...
+                    ret.Add(row);
                 }
             }
 
-            return dt;
+            return ret;
         }
     }
 
+    public sealed class ScenarioTestResultRow
```
jorive: >public [](start = 4, length = 6)

Can we make this internal?

jorive:
Need to increase the version of the API on the `BuildForPublication.cmd` script

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -78,6 +78,10 @@ public void RunBenchmarks(string assemblyFileName)
             }
         }
 
+        Action<string> OutputFileCallback = (fileName) => {
+            WriteInfoLine($"File saved to: \"{fileName}\"");
+        };
```
jorive: This can be a `static` function now.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -162,22 +191,44 @@ public void RunScenario(ScenarioConfiguration configuration, Func<ScenarioBenchm
...
+
+            WriteTableResults(new[] { scenario }, fileNameWithoutExtension, false);
+        }
+
+        public void WriteXmlResults(ScenarioBenchmark scenario, string fileNameWithoutExtension)
+        {
             var xmlFileName = $"{fileNameWithoutExtension}.xml";
-            scenarioBenchmark.Serialize(xmlFileName);
+            scenario.Serialize(xmlFileName);
             OutputFileCallback?.Invoke(xmlFileName);
```
jorive: Could you please share example xml files when you append to the Markdown table? There should still be one file per scenario.

jorive:
>         }

Do we want to make the `Write*Results` functions public. In order to minimize external dependencies and introducing breaking changes while the Api matures and gets cleaned up we have been conservative on what we expose.

---
Refers to: src/xunit.performance.api/XunitPerformanceHarness.cs:242 in e169525. [](commit_id = e169525ebf1179f4d6dd4ecc467b012b73957600, deletion_comment = False)

src/xunit.performance.api/ScenarioTest.cs
```diff
@@ -3,9 +3,9 @@
 
 namespace Microsoft.Xunit.Performance.Api
 {
-    public sealed class Scenario : IDisposable
+    public sealed class ScenarioTest : IDisposable
     {
-        public Scenario(ScenarioConfiguration configuration)
+        public ScenarioTest(ScenarioTestConfiguration configuration)
```
brianrob: @dsplaisted, can you share some context on the name change from Scenario to ScenarioTest?  I'm looking for a concrete example of the overall goal of this change.  It sounds to me, but I've not been able to confirm this in code, that the goal is to have scenario, like "Build", but have it run multiple flavors of that scenario with very minimal changes - clean, no changes.  Thus, a scenario is comprised of multiple tests, each which are slightly different.  Assuming I'm understanding things correctly here, do we need to make this change to the overall architectural model, or can we have multiple scenario benchmarks that simply run very similar code because they are similar.  I think I might be missing the reason why the existing model isn't sufficient to meet this goal.

brianrob: I should also say that if we do need this, I'm OK with extending the model - I just want to understand what's missing from the current model.

dsplaisted: @brianrob I was matching the structure of the XML result file that's generated.  The ScenarioBenchmark is the root element, and underneath it there can be multiple tests (and each test can have multiple iterations).

It seemed useful to me to be able to group the tests instead of having a flat list.  In this case, the organization I went with was to have the top level groups (scenarios) be a given project, and then the tests were different operations that would be performed on that project (such as clean build, incremental build, restore, etc).

src/xunit.performance.api/XunitPerformanceHarness.cs
dsplaisted: I'm still writing a separate XML file for each scenario, but I'm aggregating multiple scenarios for the Markdown and CSV files.  Here's the code: https://github.com/dsplaisted/sdk/blob/916ffbe41f910a2c729ef84f735e3ac3bbd5752a/test/Microsoft.NET.Perf.Tests/PerfTest.cs#L123-L133

jorive:
You will need to merge latest changes into your branch. I believe the current build number is 10.

---
In reply to: [340946867](https://github.com/Microsoft/xunit-performance/pull/243#issuecomment-340946867) [](ancestors = 340946867)

src/xunit.performance.api/ScenarioTest.cs
jorive: @Daniel Plaisted, With the current model, you can group tests under the same "Namespace" by using the name of the group it belongs to (not a very well thought design as it was copied from the xUnit micro benchmarks model).

---
In reply to: [148643709](https://github.com/Microsoft/xunit-performance/pull/243#discussion_r148643709) [](ancestors = 148643709)

jorive: I'm open to updating the actual model to something hierarchical.

---
In reply to: [148890487](https://github.com/Microsoft/xunit-performance/pull/243#discussion_r148890487) [](ancestors = 148890487,148643709)

xunit-performance.sln
```diff
@@ -0,0 +1,67 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
```
jorive: I am not sure we want to add a solution file to the project.

dsplaisted: How do you normally open the code in Visual Studio?

jorive: When you open a `csproj` file, VS automatically creates a `sln` file, and if needed I modify it locally. The reason I have not added it before is because with every VS version changes it modified the file too and I find it noisy.

jorive: @brianrob Should we add a `sln` file?

dsplaisted: If you open a csproj file directly, VS doesn't automatically add any referenced projects to the solution it creates.  So if you opened up `scenariobenchmark.csproj`, you end up having to manually add `xunit.performance.api.csproj`, `xunit.performance.core.csproj`, `xunit.performance.execution.csproj`, and `xunit.performance.metrics.csproj` to the solution to get successful build / intellisense in VS.

I think almost all of the dotnet repos have `.sln` files checked in for this reason.

dsplaisted:
@jorive 

> Do we want to make the `Write*Results` functions public. In order to minimize external dependencies and introducing breaking changes while the Api matures and gets cleaned up we have been conservative on what we expose.

This functionality needs to be public in order to support running multiple tests and then aggregating the results.

dsplaisted:
@jorive @brianrob @DrewScoggins I think I've addressed all the existing feedback

src/xunit.performance.api/ScenarioTest.cs
brianrob: Thanks @dsplaisted.  Does that mean that the way you run the benchmarks is unaffected - the change just affects the organization of test results?

xunit-performance.sln
brianrob: I don't mind adding a solution file.  If we're going to, let's keep that change as a separate commit when the rest get squashed since it's unrelated to the rest of the changes.

brianrob:
Thanks @dsplaisted, sorry for the delay on looking at this again.

dsplaisted:
> I don't mind adding a solution file. If we're going to, let's keep that change as a separate commit when the rest get squashed since it's unrelated to the rest of the changes.

@brianrob We don't normally squash commits in our repos, so I wasn't planning on doing so here.  If you prefer them squashed (except the .sln file) I'll happily do so here.

PR #244, opened 2017-10-27
Renaming the final output files (xml/csv/md) to the same name as the scenario.

jorive:


jorive:
The harness generates file names like this: `Perf-Off-CoreRun.xml` (it gets it from the process that is launched), but it should be `Perf-Off-JitBench.xml` instead (the scenario name).

DrewScoggins:
LGTM

PR #246, opened 2017-11-10
Small series of bug fixes

jorive:
- Remove an extra ; from the `<TargetFrameworks>` on the simpleharness.csproj file.
- xUnit micro-benchmarks harness: Call `GC.Collect` once more right after `GC.WaitForPendingFinalizers`
- Added more docstrings to public apis
- Updated the default Etw kernel/user providers, and remove duplication
- Minor changes in the xUnit Api tests verbosity
- Handle parsing `$CLRHELPERS$` methods generated by the Jit on x86
- Handle reloading unloaded modules

What is missing here?
- Implementation is complete and correct, but performance is not the best, it can be improved.

### DIFF
--- a/src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMonitorCounter.cs
+++ b/src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMonitorCounter.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Xunit.Performance.Api
             get
             {
                 yield return new KernelProviderInfo() {
-                    Keywords = unchecked((ulong)(KernelTraceEventParser.Keywords.PMCProfile | KernelTraceEventParser.Keywords.Profile)),
+                    Keywords = unchecked((ulong)KernelTraceEventParser.Keywords.PMCProfile),
                     StackKeywords = unchecked((ulong)KernelTraceEventParser.Keywords.PMCProfile),
                 };
                 yield return new CpuCounterInfo() {
--- a/src/xunit.performance.api/Profilers/Etw/KernelProvider.cs
+++ b/src/xunit.performance.api/Profilers/Etw/KernelProvider.cs
@@ -12,9 +12,6 @@ namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
     {
         static KernelProvider()
         {
-            // Currently, all pmc set by the API have these flags set:
-            // - Flags = PMCProfile
-            // - StackCapture = PMCProfile
             Default = new KernelProvider {
                 Flags = KernelTraceEventParser.Keywords.ImageLoad
                     | KernelTraceEventParser.Keywords.Process
--- a/src/xunit.performance.api/Profilers/Etw/KernelProvider.cs
+++ b/src/xunit.performance.api/Profilers/Etw/KernelProvider.cs
@@ -13,14 +13,14 @@ namespace Microsoft.Xunit.Performance.Api.Profilers.Etw
         static KernelProvider()
         {
             // Currently, all pmc set by the API have these flags set:
-            // - Flags = Profile | PMCProfile
+            // - Flags = PMCProfile
             // - StackCapture = PMCProfile
             Default = new KernelProvider {
                 Flags = KernelTraceEventParser.Keywords.ImageLoad
                     | KernelTraceEventParser.Keywords.Process
+                    | KernelTraceEventParser.Keywords.Profile
                     | KernelTraceEventParser.Keywords.Thread,
-                StackCapture = KernelTraceEventParser.Keywords.ContextSwitch
-                    | KernelTraceEventParser.Keywords.SystemCall,
+                StackCapture = KernelTraceEventParser.Keywords.None,
             };
         }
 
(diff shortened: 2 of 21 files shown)


src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMonitorCounter.cs
```diff
@@ -17,7 +17,7 @@ public override IEnumerable<ProviderInfo> ProviderInfo
             get
             {
                 yield return new KernelProviderInfo() {
-                    Keywords = unchecked((ulong)(KernelTraceEventParser.Keywords.PMCProfile | KernelTraceEventParser.Keywords.Profile)),
+                    Keywords = unchecked((ulong)KernelTraceEventParser.Keywords.PMCProfile),
```
DrewScoggins: I assume that PMCProfile was already setting all of the bits that Profile was? Or is this a functional change?

jorive: `PMCProfile` is not a default anymore, but only set when Instructions Retired, Cache Misses or Branch Mispredictions are specified.

jorive: `Profile` is a default flag (always captured). I removed some of the redundancy here, and flags that were set but not used.

src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -32,53 +39,72 @@ public XunitPerformanceHarness(string[] args)
...
+            }
+
             var xUnitPerformanceSessionData = new XUnitPerformanceSessionData {
                 AssemblyFileName = assemblyFileName,
-                CollectOutputFilesCallback = (fileName) => {
-                    // FIXME: This will need safe guards when the client calls RunBenchmarks in different threads.
-                    _outputFiles.Add(fileName);
-                    WriteInfoLine($"File saved to: \"{fileName}\"");
-                },
+                CollectOutputFilesCallback = LogFileSaved,
```
DrewScoggins: It looks like we no longer add fileName to _outputFiles. This had a FIXME attached to it, is this fixed? I did not see anything else in the diff that obviously fixed this.

src/xunit.performance.api/Profilers/Etw/KernelProvider.cs
```diff
@@ -13,14 +13,14 @@ internal sealed class KernelProvider
         static KernelProvider()
         {
             // Currently, all pmc set by the API have these flags set:
-            // - Flags = Profile | PMCProfile
+            // - Flags = PMCProfile
             // - StackCapture = PMCProfile
```
AndyAyersMS: Comment it stale here

src/xunit.performance.api/XunitPerformanceHarness.cs
jorive: It is not used, so I removed it. We only use this function to display generated files.

src/xunit.performance.api/Profilers/Etw/UserProvider.cs
```diff
@@ -24,7 +24,11 @@ static UserProvider()
                 },
                 new UserProvider {
                     Guid = ClrTraceEventParser.ProviderGuid,
-                    Keywords = (ulong)(ClrTraceEventParser.Keywords.Exception | ClrTraceEventParser.Keywords.GC | ClrTraceEventParser.Keywords.Jit | ClrTraceEventParser.Keywords.JittedMethodILToNativeMap | ClrTraceEventParser.Keywords.Loader),
+                    Keywords = (ulong)(ClrTraceEventParser.Keywords.Exception
```
DrewScoggins: Why did we add the NGen flag and remove the JittedMethodILToNativeMap flag?

jorive: It was a suggestion from @AndyAyersMS. We are looking at ngen data, and currently we were setting "JittedMethodILToNativeMap" and not really using it.

src/xunit.performance.api/PerformanceMonitorCounter/BasePerformanceMonitorCounter.cs
DrewScoggins: Yes, I understand.

src/xunit.performance.api/Profilers/Etw/UserProvider.cs
jorive: If you think we need it, then I can add it back. I was just trimming flags that we are not using, and makes the generated etl bigger.

DrewScoggins: OK, sounds good.

src/xunit.performance.api/XunitPerformanceHarness.cs
DrewScoggins: Alright, sounds good to me.

src/xunit.performance.api/Profilers/Etw/UserProvider.cs
AndyAyersMS: We don't need the IL to native maps for our current analysis.

jorive:
@AndyAyersMS I agree, and I also wanted to revisit the flags we capture and we use, so I could remove the unnecessary ones to reduce the runtime overhead and generated file size. We were capturing a lot of data we were not parsing/using.

PR #248, opened 2017-11-16
System.Diagnostics.Process has a different behavior on non-Windows platforms

jorive:
- The `Process.StartTime` throw `System.ComponentModel.Win32Exception` on Linux because the data is not available once the process exits.

### DIFF
--- a/src/xunit.performance.api/ProcessExitInfo.cs
+++ b/src/xunit.performance.api/ProcessExitInfo.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Diagnostics;
-using System.Runtime.InteropServices;
 
 namespace Microsoft.Xunit.Performance.Api
 {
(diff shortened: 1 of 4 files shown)


src/xunit.performance.api/ProcessExitInfo.cs
```diff
@@ -11,18 +12,31 @@ public sealed class ProcessExitInfo
...
                 throw new InvalidOperationException($"{process.ProcessName} has not exited.");
+            if (exitTime < startTime)
+                throw new InvalidOperationException($"Process.ExitTime: {exitTime}, is less than the Process.StartTime: {startTime}.");
 
             ProcessId = process.Id;
             ExitCode = process.ExitCode;
-            StartTime = process.StartTime;
-            ExitTime = process.ExitTime;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
brianrob: The change here seems reasonable to me, but I think rather than special casing Windows, just do the same thing on all platforms.  Otherwise, our methodology is different between operating systems, and that makes both the harness and data interpretation more complicated.

jorive: Thanks, makes sense.

Issue #220, opened 2017-04-18
Could not load file or assembly 'xunit.runner.utility.dotnet, Version=2.2.0.3300


deviantdes:
Hi guys, ive created a new .NET Framework 4.6.2 project and have been trying to get this xunit.performance.api to work (I have followed the steps in the README, and i actually got it to work in a test .NET Core project. However my work requires me to use .NET Framework)

Sadly for the .NET Framework project, i face this error when i try to run 

> System.IO.FileNotFoundException: 'Could not load file or assembly 'xunit.runner.utility.dotnet, Version=2.2.0.3300, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c' or one of its dependencies. The system cannot find the file specified.'

Is it a configuration issue ? 

installed version:
xunit.performance.api 1.0.0-beta-build0001

The dependencies for this version is .NETStandard v1.5 so i assume that .NET Framework 4.6.2 might work.. or am i wrong

Has anyone had success with .NET Framework ?

Thanks in advance guys.

jorive:
@deviantdes Could you please try again with the latest version of the API (build0007)? I believe this was fix in previous versions.
Thanks,
-José

mycroes:
I have the same exception, with xunit.performance.api 1.0.0-beta-build0013. The target framework for my application is .NET 4.6.1.

jorive:
@mycroes with #251 this issue is fixed. Please feel free to open a new issue if you are still running into problems.

PR #251, opened 2017-12-01
Creating new Desktop binaries to unblock Desktop support.

jorive:
Our packages were targeting netstandard1.5, but they are not working when targeting desktop. While we are investigating, I'm adding to the build matrix net461 to unblock desktop perf investigations.


jorive:
This also fixes this issue: https://github.com/Microsoft/xunit-performance/issues/220

jorive:
@nategraf This PR enables you to run on Desktop

PR #252, opened 2017-12-21
Minor changes and bug fixes.

jorive:
- After running all benchmarks now we throw an exception if a benchmark failed.
  - This is mainly meant to catch bad code gen or broken benchmarks on automation.
- Add an example where a Benchmark attribute is marked to be skipped. 
- Disable the GetAllocatedBytesForCurrentThreadTest on non-netcoreapp[1.1|2.0]
- Updating some error message strings. 
- Simplified a DEBUG ONLY function.

### DIFF
--- a/src/xunit.performance.api/XunitPerformanceHarness.cs
+++ b/src/xunit.performance.api/XunitPerformanceHarness.cs
@@ -68,7 +68,9 @@ namespace Microsoft.Xunit.Performance.Api
 
             void xUnitAction(string assemblyPath)
             {
-                XunitRunner.Run(assemblyPath, _typeNames);
+                var errCode = XunitRunner.Run(assemblyPath, _typeNames);
+                if (errCode != 0)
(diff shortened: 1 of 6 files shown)


src/xunit.performance.api/XunitPerformanceHarness.cs
```diff
@@ -68,7 +68,9 @@ public void RunBenchmarks(string assemblyFileName)
 
             void xUnitAction(string assemblyPath)
             {
-                XunitRunner.Run(assemblyPath, _typeNames);
+                var errCode = XunitRunner.Run(assemblyPath, _typeNames);
+                if (errCode != 0)
+                    throw new Exception($"{errCode} benchmark(s) did not complete successfully.");
```
brianrob: I'd recommend making this text a bit more clear.  I think you're trying to say that some number of benchmarks actually failed to execute (perhaps threw an exception).  We should try to make clear that these tests failed to execute - not that the performance of the tests was not acceptable.

PR #257, opened 2018-01-16
Update xUnit and CommandLineParser NuGet packages to stable versions

alexanderkozlenko:
- Update xUnit packages to version 2.2.0
- Update CommandLineParser package to version 2.2.1

### DIFF
--- a/src/xunit.performance.api/PerformanceTestMessageSink.cs
+++ b/src/xunit.performance.api/PerformanceTestMessageSink.cs
@@ -8,21 +8,22 @@ using Xunit.Abstractions;
 
 namespace Microsoft.Xunit.Performance.Api
 {
-    internal sealed class PerformanceTestMessageVisitor : TestMessageVisitor<IDiscoveryCompleteMessage>
+    internal sealed class PerformanceTestMessageSink : TestMessageSink
     {
-        public PerformanceTestMessageVisitor()
+        public PerformanceTestMessageSink()
         {
             Tests = new List<PerformanceTestMessage>();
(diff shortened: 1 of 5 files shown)


jorive:
@alexanderkozlenko There are not benchmark results with these changes. This is from the logs:
```md
 simpleharness.exe | Metric | Unit | Iterations | Average | STDEV.S | Min | Max
:----------------- |:------ |:----:|:----------:| -------:| -------:| ---:| ---:
```
Could you please take a look?

alexanderkozlenko:
Fixed with b6e3d0c96264893c6a8b7353328e278dcd6a079c

src/xunit.performance.api/PerformanceTestMessageSink.cs
```diff
@@ -78,5 +87,15 @@ private static Type GetType(string assemblyName, string typeName)
...
+
+        public override void Dispose()
+        {
+            Discovery.DiscoveryCompleteMessageEvent -= OnDiscoveryCompleteMessageEvent;
+            Discovery.TestCaseDiscoveryMessageEvent -= OnTestCaseDiscoveryMessageEvent;
+
+            base.Dispose();
+        }
+
+        public ManualResetEvent Finished => _finished;
```
jorive: Could you please remove the private member variable `_finished`, and use property instead (like the `Test` property above):
```csharp
...
public PerformanceTestMessageSink()
{
    Finished = new ManualResetEvent(false);
    Tests = new List<PerformanceTestMessage>();
...
public ManualResetEvent Finished { get; }
```

PR #259, opened 2018-02-15
Upgrade to TraceEvent 2.0.5

brianrob:
Updated: This PR updates xunit-performance to use TraceEvent 2.0.5 which contains the fixes necessary to run against netstandard1.6.

------------------------------
OLD:

In order to move off of the pre-release version of TraceEvent to a shipped version (2.0.4) we need to do two this:

1. Update the nuget package reference to 2.0.4.
2. Remove support for netcoreapp1.0 and 1.1.

Older versions of the harness can still run on netcoreapp1.0 and 1.1, but the official TraceEvent package does not support versions of netcoreapp older than 2.0.  This is because there are some support DLLs that won't run on previous versions without a decent amount of work.  They work on 2.0 because enough type-forwarders have been added to mscorlib to handle the work that these DLLs do, but these type-forwarders don't exist pre-netcoreapp2.0.

Per my conversation with @valenis, we do not need to continue to produce new harnesses that support pre-netcoreapp2.0, and thus rather than attempting to do the work to make the new version of TraceEvent support netcoreapp1.0 and 1.1, we can drop support for them here.

### DIFF
--- a/src/xunit.performance.api/xunit.performance.api.csproj
+++ b/src/xunit.performance.api/xunit.performance.api.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.api</AssemblyTitle>
-    <TargetFrameworks>netstandard1.5;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
     <Title>xUnit Performance Api</Title>
     <RootNamespace>Microsoft.Xunit.Performance.Api</RootNamespace>
@@ -18,6 +18,7 @@
     </PackageReference>
     <PackageReference Include="System.ValueTuple" Version="4.3.1" />
     <PackageReference Include="System.Xml.XmlSerializer" Version="4.3.0" />
+    <PackageReference Include="System.Diagnostics.Process" Version="4.3.0" />
     <PackageReference Include="xunit.runner.utility" Version="$(XunitPackageVersion)" />
   </ItemGroup>
 
--- a/src/xunit.performance.api/xunit.performance.api.csproj
+++ b/src/xunit.performance.api/xunit.performance.api.csproj
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.api</AssemblyTitle>
     <TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.6</TargetFrameworks>
     <Title>xUnit Performance Api</Title>
     <RootNamespace>Microsoft.Xunit.Performance.Api</RootNamespace>
   </PropertyGroup>
--- a/src/xunit.performance.metrics/xunit.performance.metrics.csproj
+++ b/src/xunit.performance.metrics/xunit.performance.metrics.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.metrics</AssemblyTitle>
-    <TargetFrameworks>netstandard1.5;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
     <Title>xUnit Performance Metrics</Title>
   </PropertyGroup>
--- a/src/xunit.performance.metrics/xunit.performance.metrics.csproj
+++ b/src/xunit.performance.metrics/xunit.performance.metrics.csproj
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.metrics</AssemblyTitle>
     <TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.6</TargetFrameworks>
     <Title>xUnit Performance Metrics</Title>
   </PropertyGroup>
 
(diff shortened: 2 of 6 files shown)


jorive:
[Microsoft.Diagnostics.Tracing.TraceEvent](https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/) supports `netstandard1.6`, so it should work on `netcoreapp1.x`. It sounds to me like a bug on the TraceEvent library. Have we verified it?

src/xunit.performance.api/xunit.performance.api.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.api</AssemblyTitle>
-    <TargetFrameworks>netstandard1.5;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
```
jorive: This line should be updated for non-Windows platforms too

src/xunit.performance.metrics/xunit.performance.metrics.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>xunit.performance.metrics</AssemblyTitle>
-    <TargetFrameworks>netstandard1.5;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
```
jorive: Same as above.

tests/scenariobenchmark/scenariobenchmark.csproj
```diff
@@ -6,7 +6,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp1.0;netcoreapp1.1;netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp1.0;netcoreapp1.1;netcoreapp2.0</TargetFrameworks>
```
jorive: Same as above.

tests/simpleharness/simpleharness.csproj
```diff
@@ -6,7 +6,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp1.0;netcoreapp1.1;netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp1.0;netcoreapp1.1;netcoreapp2.0</TargetFrameworks>
```
jorive: Same as above.

tests/standalonesample/standalonesample.csproj
```diff
@@ -6,12 +6,12 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp1.0;netcoreapp1.1;netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp1.0;netcoreapp1.1;netcoreapp2.0</TargetFrameworks>
```
jorive: Same as above.

tests/standalonesample/standalonesample.csproj
```diff
@@ -6,12 +6,12 @@
...
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.3-alpha-experimental">
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.4">
       <IncludeAssets>All</IncludeAssets>
```
jorive: If the package was fixed on TraceEvent, then IncludeAsset/All is not needed.

src/xunit.performance.api/xunit.performance.api.csproj
```diff
@@ -18,6 +18,7 @@
     </PackageReference>
```
jorive: If the package was fixed on TraceEvent, then IncludeAsset/All is not needed.

jorive:
`README.md` might need to be updated.

brianrob:
Yes, I have confirmed what the issue is in the debugger and with ildasm.

TraceEvent depends on a SupportFiles package which contains binaries that aren't open source.  Two of them are TLBImp binaries - managed wrappers around COM objects.  The third is OSExtensions.dll which contains non-OSS operating system specific code.  While the package states that it supports netstandard1.6, the binaries themselves are targeted at desktop (even the ones that are in the netstandard1.6 directory).  In order to fix this, we would need to generate new netstandard TLBImp binaries, which it does not seem possible to do, and then re-generate a netstandard1.6 OSExtensions.dll and post new packages.  There are some hacks that we may be able to do to address this, but it is not clear that they will work.  We could also consider just re-generating OSExtensions.dll, which would fix the problem we are running into here, but wouldn't necessarily get used since it only affects older versions of core.

Thus the reason that I propose dropping support for netcoreapp1.0 and 1.1.  I have also discussed the TraceEvent specifics with @vancem and he is OK as long as the functionality works for netcoreapp2.0 and later.

jorive:
`TraceEvent` then needs to update what it targets because this is misleading.

jorive:
If we are moving with 2.0 only, then we can remove the need to download the other SDKs on `dotnet-install.cmd` too.

brianrob:
To be clear, the failure is understood.  There are desktop assemblies being used on netstandard1.6 and there is a decent amount of work to fix that within TraceEvent.  The reason it works on netcoreapp2.0 is because the APIs that are needed happen to exist as TypeForwarders in mscorlib.  This was done as part of netcoreapp2.0 for compatibility  So, you are right, this is a workaround for an issue in TraceEvent.  I am looking more at what it will take to fix TraceEvent, but the answer might be to just mark TraceEvent as netstandard2.0 and move on in which case this PR will still need to move forward.  I suspect that there is a decent amount of work to fix TraceEvent to make it netstandard1.6 compliant, but we shall see.

My goal right now is to get Xunit-Performance in a state where it can use an official TraceEvent so that we can enable ETW testing on CoreFX, which we can't do without this change.

brianrob:
Per-offline conversation, I have done the work to fix TraceEvent, and have published TraceEvent 2.0.5.  This PR consumes the new package and preserves support for netcoreapp1.0 and netcoreapp1.1.

PR #264, opened 2018-03-09
NIT: rename timespan to timeoutPerIteration to make clear what it is for the caller

adamsitnik:


jorive:
Do you want a new package? If so, could you please update the package version?

adamsitnik:
@jorive no, it's more a cosmetic change that can be shipped with some future version

PR #270, opened 2018-07-23
Remove unstable package & cross compile executor

ViktorHofer:
@jorive this fixes the test error and also replaces the alpha dependency with a stable netstandard2.0 package. This removes a compiler warnings because of two different assembly identities. I also opened a PR but I doubt it will be accepted as the project seems dead: https://github.com/Wheelies/MarkdownLog/pull/16

You can now publish the packages. Thanks!


ViktorHofer:
@dotnet-bot test this please

PR #271, opened 2018-07-23
Update .NET cli verion used to build xUnit-Performance

jorive:


jorive:
@ViktorHofer PTAL

.vscode/extensions.json
```diff
@@ -1,5 +1,7 @@
 {
     "recommendations": [
-        "ms-vscode.csharp"
+        "ms-vscode.csharp",
+        "DavidAnson.vscode-markdownlint",
+        "ms-vscode.csharp",
```
ViktorHofer: usually we don't include the .vscode directory in git

DotNetCliVersion.txt
```diff
@@ -1 +1 @@
-2.0.0
\ No newline at end of file
+2.1.302
```
ViktorHofer: this value is not picked up by the sh script right? As you hardcoded it there.

dotnet-install.cmd
```diff
@@ -30,25 +28,24 @@ setlocal
...
   if not exist "%DotNet_Path%\dotnet-install.ps1" (
     call :print_error_message Failed to download "%DotNet_Path%\dotnet-install.ps1"
     exit /b 1
   )
 
-  echo Executing dotnet installer script "%DotNet_Path%\dotnet-install.ps1"
-  for %%v in (2.0.0) do (
-    echo Installing dotnet sdk version %%~v
+  call :print_header_message Executing dotnet installer script "%DotNet_Path%\dotnet-install.ps1"
+  for %%v in (%DotNet_Version%) do (
```
ViktorHofer: we can get rid of the loop here, right?

ViktorHofer (approved):
LGTM if it builds well.

.vscode/extensions.json
jorive: This file provides suggestions of the extensions to download. I think it is useful, specially for new users.

DotNetCliVersion.txt
jorive: This version is used by the `.cmd` and `.sh` files. Ideally, I would consolidate the build scripts into a single script.

dotnet-install.cmd
jorive: Sure.

DotNetCliVersion.txt
ViktorHofer: So is the hardcoded version string here needed? https://github.com/Microsoft/xunit-performance/pull/271/files#diff-ba04653bd26665aa2b926dbe2b144e61R21

.vscode/extensions.json
ViktorHofer: fine with me :)

DotNetCliVersion.txt
jorive: Good call, it needs to be updated.

ViktorHofer:
No idea why GitHub says that I authored your first commit here 😂. LGTM. I'm happy that repo is now finally again up2date code and infrastructure wise :)

jorive:
Me neither... I'm confused :P

ViktorHofer:
Maybe incrementing the versions to 1.0.0 would really make sense. Not sure...

jorive:
Let's do it with the next source code changes. Maybe adding Code Analysis?

ViktorHofer:
Oooh that sounds fancy 😁

